PR #33, opened 2016-09-27
Initial browser modeling r? @silne30

jrbenny35:
Just getting this started so there is some idea of what is going on. I have a few tests and Docs started but nothing really done with them. The browser model works but it needs a lot of assistance from the base_window and windows classes. Of course we can add more to those as we move on.


silne30:
@jrbenny35 This looks really promising. I know this is going to sound so petty but for some reason the Jenkins job failed on the flake8. It literally failed because of being one character over. Can you fix that so that your tests can run and then we can continue with the work you were doing?

foxpuppet/tests/conftest.py
```diff
@@ -0,0 +1,15 @@
+import pytest
+from foxpuppet.foxpuppet import FoxPuppet
+from foxpuppet.ui.browser.window import Browser_Window
```
whimboo: Do we need the 'ui' part in the module hierarchy? I thought that foxpuppet should only care about ui.

foxpuppet/tests/conftest.py
```diff
@@ -0,0 +1,15 @@
+import pytest
+from foxpuppet.foxpuppet import FoxPuppet
```
whimboo: Maybe make it available via __init__.py so that a 'from foxpuppet import FoxPuppet' would be enough.

foxpuppet/ui/browser/window.py
```diff
@@ -0,0 +1,90 @@
...
+        self.selenium = foxpuppet.selenium
+        self._navbar = None
+        self._tabbar = None
+
+    @property
+    def navbar(self):
+        self.selenium.switch_to()
+
+        if not self._navbar:
+            navbar = self.selenium.find_element_by_id('nav-bar')
```
whimboo: For other page object models (like for addons) I have seen that element ids have been set as constants at the top. I feel that this is better and gives faster access in case of ui changes in Firefox.

foxpuppet/ui/browser/window.py
```diff
@@ -0,0 +1,90 @@
...
+            self._navbar = Navbar(lambda: self.selenium, self, navbar)
+
+        return self._navbar
+
+    @property
+    def tabbar(self):
+        self.selenium.switch_to.window(self.windows.current)
```
whimboo: Why does it differ from the navbar implementation?

foxpuppet/ui/browser/window.py
```diff
@@ -0,0 +1,90 @@
...
+
+        if not self._tabbar:
+            tabbrowser = self.selenium.find_element_by_id('tabbrowser-tabs')
+            self._tabbar = TabBar(lambda: self.selenium, self, tabbrowser)
+
+        return self._tabbar
+
+    def close(self, handle):
+        self.selenium.switch_to.window(handle)
```
whimboo: So we want to allow to close different windows here? Shouldn't it operate on eg. self.handle?

foxpuppet/ui/browser/window.py
```diff
@@ -0,0 +1,90 @@
...
+        self.selenium.find_element_by_id('titlebar-close').click()
```
whimboo: Separating out element retrieval and actions? It would give re-usable code, in case a test only wants to have the close button element.

foxpuppet/ui/browser/window.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def close_all(self, exceptions=None):
```
whimboo: Close all what? From the top browser level I don't get what we are referring here. I find it confusing.

foxpuppet/ui/browser/window.py
```diff
@@ -0,0 +1,90 @@
...
+
+        windows_to_keep = exceptions or []
+
+        for handle in self.windows.all:
+            if windows_to_keep == handle:
+                continue
+            self.selenium.switch_to.window(handle)
+            self.close(handle)
+
+    def min_window_size(self):
```
whimboo: Confusing method name for me. Why not 'minimize()'? Same for maximize().

foxpuppet/ui/browser/window.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def max_window_size(self):
+        self.selenium.switch_to.window(self.windows.current)
+        button = self.selenium.find_element_by_id('titlebar-max')
+
+        button.click()
+
+    def new_private_browsing_window(self):
+        self.selenium.switch_to.window(self.windows.current)
+        self.selenium.find_element_by_id('PanelUI-menu-button').click()
```
whimboo: I feel an 'open_browser()' method would be more accurate, similar to what we have in puppeteer. It will allow you to use alternatives in how the window gets opened. That is what we also want to have.

foxpuppet/tests/conftest.py
jrbenny35: Ahh this is good thinking. Learned something new today.

foxpuppet/ui/browser/window.py
silne30: @whimboo Should we have a flag for whether the the window will be private or not?

jrbenny35: I think we should. I am thinking open_window(self, private=None) and have that flag of private added to the window object for that instance.

foxpuppet/ui/browser/window.py
whimboo: CC'ing @davehunt here for his feedback.

foxpuppet/ui/browser/window.py
whimboo: That's the implementation we currently have for puppeteer:

https://dxr.mozilla.org/mozilla-central/source/testing/puppeteer/firefox/firefox_puppeteer/ui/browser/window.py#185

It uses the flag twice. One time for opening the window via the menu entry, and another time via a shortcut.

foxpuppet/tests/conftest.py
silne30: @jrbenny35 I think this a a good point. Let's remove ui from the heirarchy.

.travis.yml
```diff
@@ -24,4 +24,4 @@ before_script:
   - "flake8 ."
 
 script:
-  - "py.test --capture=sys --showlocals --tb=long foxpuppet/tests "
\ No newline at end of file
+  - "py.test --capture=sys --showlocals --tb=long --driver Firefox foxpuppet/tests "
```
davehunt: Add a new line to the end of the file.

davehunt: `--capture=sys`, `--showlocals` and `--tb-long` can all be useful options when debugging failures, but I don't think we need them for every Travis CI build.

foxpuppet/tests/conftest.py
```diff
@@ -0,0 +1,15 @@
+import pytest
```
davehunt: Let's include the license block in all `.py` files.

foxpuppet/__init__.py
```diff
@@ -1,3 +1,9 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
+
+
```
davehunt: Nit: We can remove one of these empty lines.

foxpuppet/tests/conftest.py
```diff
@@ -0,0 +1,15 @@
...
+from foxpuppet import FoxPuppet
+from foxpuppet.ui.browser.window import Browser_Window
+
+
+@pytest.fixture
+def set_up(selenium):
```
davehunt: Let's call this fixture `foxpuppet` as that's what it's returning.

foxpuppet/tests/test_browser_model.py
```diff
@@ -0,0 +1,57 @@
...
+
+    # Tests minimizing the browser window
+    def test_minimize_browser_window(self, windows, selenium):
+
+        FoxPuppet(selenium)
+        selenium.get('http://www.google.com')
+        selenium.set_context('chrome')
+        windows.min_window_size()
+        windows.close_all()
+        selenium.quit()
```
davehunt: There's no need to call `selenium.quit()` as pytest-selenium will do that for you. I think we can simplify this by using a `foxpuppet` fixture. Personally I'd prefer the test to look something like:

```python
def test_minimize_window(foxpuppet):
    foxpuppet.window.minimize()
    # assert something
```

There's no need to open a website (this also makes our tests dependent on that website being available, and having a network connection). I also think that `FoxPuppet` should provide access to the window directly rather than having a separate `Browser_Window` object to instantiate. Finally, the switching of context should be masked from the user and handled in the FoxPuppet methods.

foxpuppet/tests/conftest.py
davehunt: Alternatively, we could call it `firefox` as that's what it represents.

foxpuppet/tests/test_browser_model.py
```diff
@@ -0,0 +1,57 @@
+from foxpuppet import FoxPuppet
+
+
+class Test_Browser_Window:
```
davehunt: No need for a test class.

foxpuppet/ui/browser/window.py
```diff
@@ -0,0 +1,90 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from .tabbar import TabBar
+from .navbar import Navbar
+from foxpuppet.ui.windows import Windows
+from foxpuppet.ui.base_window import BaseWindow
+
+
+class Browser_Window(object):
```
davehunt: Class names should be named as `BrowserWindow` though I suspect 'Browser' is redundant in our case

foxpuppet/ui/browser/window.py
davehunt: Totally agree with @whimboo.

foxpuppet/ui/browser/window.py
davehunt: This is not typical for web projects, but I can see value if an element is used in more than one place. One suggestion would be to make the properties that return elements to be private, so it's clear they are to be used internal to the UI objects.

foxpuppet/ui/browser/window.py
davehunt: I actually like matching the menu labels here, which is what you have currently.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 selenium == 3.0.0b2
 flake8
 pytest == 2.9.2
+pytest-selenium
```
davehunt: pytest and pytest-selenium are test requirements, and should not be installed as dependencies for FoxPuppet. I would recommend listing FoxPuppet dependencies in `setup.py` and using Tox to run your tests, listing test dependencies in `tox.ini`.

foxpuppet/ui/browser/window.py
jrbenny35: From exploring the browser chrome, Navbar is where the search bar is and Tabbar is where the tabs are. That's the only reason why I separated them

foxpuppet/ui/browser/window.py
jrbenny35: I have no honest idea on how to keep track of handles. There just a string of random numbers it seems. I could use some help with handle tracking.

foxpuppet/ui/browser/window.py
whimboo: We will have to handle different types of windows in case of chrome scope. So removing browser could be misleading.

foxpuppet/ui/browser/window.py
whimboo: That's not what I'm referring to. The question is more why you need the `handle` option here. This method should only work on the handle as the window has been created for. See our Firefox puppeteer implementation.

Btw. the outer window id (window handle) changes if the remoteness of a tab changes (only for electrolysis). It only affects tabs but not chrome windows. See also https://bugzilla.mozilla.org/show_bug.cgi?id=1259055#c17 for a related discussion.

foxpuppet/ui/browser/window.py
davehunt: What other types of window are you referring to? I know that 'window' can refer to many things, but I imagine a user of FoxPuppet would probably only use it to refer to the browser window.

whimboo: If we want to use FoxPuppet for chrome, we have dozen of other windows like library, about, page info, ...

davehunt: I see, that makes sense then. 👍

silne30:
Closed in favor of #34

Issue #25, opened 2016-09-16
Documentation


silne30:
This epic is for the continual grooming and development of the docs for this project.

jrbenny35:
@davehunt I'd like to make some progress on this. Is there a good example of what I should be aiming for?

davehunt:
This issue is too vague, let's close it and work on more specific issues such writing a user guide and generating API documentation from the code.

Issue #24, opened 2016-09-16
Generate API documentation from source code


silne30:
We need to generate the API documentation from the source using docstrings. For this issue we should create an API page in the documentation that includes all public classes and methods with docstrings. We can then follow up with improving the structure and docstrings.

davehunt:
@silne30 please open issues with a description as well as a title, otherwise it's hard for contributors to know what they need to work on or how to get started.

davehunt:
You can see an example here: http://pypom.readthedocs.io/en/latest/api.html

davehunt:
Landed in #67

Issue #23, opened 2016-09-16
Configure ReadTheDocs


silne30:

davehunt:
This is done, our documentation is built whenever we push, and hosted at http://foxpuppet.readthedocs.io/

PR #34, opened 2016-09-30
Additional travis config r? @davehunt, @silne30

jrbenny35:
Some changes to allow Tox to control the virtual environment and for geckodriver to be installed in the path.

Since we haven't really made any changes to master, this PR includes all of the changes from my Initial Browser Modeling PR as well. Please ignore those commits.


jrbenny35:
These tests are pretty simple but are proving to be flakey. If my py.test command is actually using the firefox browser it downloads, then something is wrong. 

I am thinking that either the firefox version is old, or the waits are too short. Selenium should wait until the page loads the url, so that's why I am thinking the firefox nightly isn't being used.

Any insight would be appreciated. Travis is still new to me, as I've never tested the browser chrome on travis.

.travis.yml
```diff
@@ -6,13 +6,18 @@ python:
 
 sudo: "required"
 
-before_install:
-  - "pip install virtualenv"
-  - "virtualenv venv"
-  - "source venv/bin/activate"
+env:
+  matrix:
+  - TOXENV=foxpuppet GUI=1
```
davehunt: No need for a matrix if we only have one entry. I've suggested having py27 and py35 as test environments, so let's have those two items in the matrix.

tox.ini
```diff
@@ -0,0 +1,9 @@
+[tox]
+envlist = foxpuppet
```
davehunt: I would make the envlist `py{27,35}` so that we run the tests against python 2.7 and 3.5. Later I would also introduce a `flake8` environment, but let's keep it to the tests for now.

tox.ini
```diff
@@ -0,0 +1,9 @@
...
+
+[testenv:foxpuppet]
+deps =
+	pytest==3.0.0
```
davehunt: Latest is 3.0.3, also (just a nit) the indent seems excessive here - four spaces would be better.

tox.ini
```diff
@@ -0,0 +1,9 @@
...
+	pytest-selenium==1.4.0
+	selenium==3.0.0b2
+commands = py.test --capture=sys --showlocals --tb=long --driver Firefox foxpuppet/tests {posargs}
```
davehunt: Use `pytest` instead of `py.test`

tox.ini
```diff
@@ -0,0 +1,9 @@
+[tox]
+envlist = foxpuppet
+
+[testenv:foxpuppet]
```
davehunt: Remove `:foxpuppet` and make this the defaults for the test environments.

tox.ini
davehunt: Also, remove `--capture=sys --showlocals --tb=long`

davehunt: Move `foxpuppet/tests` to a new section at the bottom:

```ini
[pytest]
testpaths = foxpuppet/tests
```

.travis.yml
```diff
@@ -6,13 +6,18 @@ python:
...
 
-install:
-  - "pip install -r requirements.txt "
+before_install:
+  - pip install mozdownload
+  - mozdownload --type=daily
```
davehunt: I would add the branch to make it clear this is a Nightly build, you can also specify a destination filename. You can also use mozinstall and symbolic links to make this the default Firefox version:

```
pip install mozdownload mozinstall
mozdownload --type daily --branch mozilla-central --destination firefox.tar.bz2
mozinstall firefox.tar.bz2
sudo ln -sf $PWD/firefox/firefox /usr/local/bin/firefox
```

I suspect that might help your tests as you'll be using the correct version of Firefox.

.travis.yml
```diff
@@ -6,13 +6,18 @@ python:
...
+  - for f in *; do if [[ $f =~ (mozilla-central|firefox-52) ]]; then tar -xvf $f; fi; break; done
+  - if [ $TOXENV ]; then sudo pip install tox; fi
```
davehunt: You shouldn't need `sudo` for installing Tox. Also, no need for the `if` as we're only using Tox.

.travis.yml
```diff
@@ -6,13 +6,18 @@ python:
...
+  - if [[ $TOXENV == 'foxpuppet' ]]; then wget -O geckodriver.tar.gz https://github.com/mozilla/geckodriver/releases/download/v0.10.0/geckodriver-v0.10.0-linux64.tar.gz; fi
+  - if [[ $TOXENV == 'foxpuppet' ]]; then gunzip -c geckodriver.tar.gz | tar xopf -; fi
+  - if [[ $TOXENV == 'foxpuppet' ]]; then chmod +x geckodriver && sudo mv geckodriver /usr/local/bin; fi
```
davehunt: Remove all the `if` blocks because we'll always need GeckoDriver. We might not want it when we're running flake8 but I'd take a more readable configuration over the cost of downloading a small binary.

.travis.yml
```diff
@@ -22,6 +27,7 @@ notifications:
 
 before_script:
   - "flake8 ."
```
davehunt: Remove this, we'll make it a Tox environment.

.travis.yml
```diff
@@ -22,6 +27,7 @@ notifications:
...
+  - if [ $GUI ]; then export DISPLAY=:99.0 && sh -e /etc/init.d/xvfb start -screen 0 1920x1080x24 && sleep 3; fi
```
davehunt: This just needs to be `sh -e /etc/init.d/xvfb start` and the `DISPLAY` can go as a global environment variable.

.travis.yml
```diff
@@ -22,6 +27,7 @@ notifications:
...
 
 script:
-  - "py.test --capture=sys --showlocals --tb=long foxpuppet/tests "
\ No newline at end of file
+  - "py.test --capture=sys --showlocals --tb=long --driver Firefox --firefox-path=firefox/firefox-bin foxpuppet/tests"
```
davehunt: We should be using Tox here 😆 so change this to simply `tox`. No need for the path to Firefox as we've made sure it's in the path.

tox.ini
```diff
@@ -0,0 +1,14 @@
...
+skipsdist = true
+envlist = py{27,35}
+
+[testenv]
+passenv = DISPLAY
+deps =
+	pytest==3.0.3
+	pytest-selenium==1.4.0
+	selenium==3.0.0b2
+commands = pytest --driver Firefox foxpuppet/tests
```
whimboo: Hm, now {posargs} is gone. Was that expected? Also the test path appears twice now. I think that this not necessary, or?

foxpuppet/tests/test_browser_model.py
```diff
@@ -0,0 +1,42 @@
...
+# Tests adding a new bookmark
+def test_bookmark_button(selenium):
+
+    foxpuppet = FoxPuppet(selenium)
+    selenium.set_context('chrome')
+    foxpuppet.bookmark_page()
+
+
+def test_new_tab_button(selenium):
+
```
whimboo: This empty new line in each method is kinda strange. Is that somehow part of your style guides?

foxpuppet/tests/test_browser_model.py
```diff
@@ -0,0 +1,42 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+
+from foxpuppet import FoxPuppet
+
+
+# Tests opening a new private browsing window via menu
```
whimboo: Maybe better making this comment a doc comment?

foxpuppet/browser/window.py
```diff
@@ -0,0 +1,96 @@
...
+from foxpuppet.browser.windows import Windows
+
+NAV_BAR = 'nav-bar'
+TAB_BROWSER = 'tabbrowser-tabs'
+TITLE_BAR_CLOSE = 'titlebar-close'
+TITLE_BAR_MIN = 'titlebar-min'
+TITLE_BAR_MAX = 'titlebar-max'
+
+
+class Window(object):
```
whimboo: So I thought we wanted to keep `BrowserWindow`? @davehunt?

tox.ini
davehunt: Yeah, let's replace `foxpuppet/tests` with `{posargs}`.

foxpuppet/tests/test_browser_model.py
davehunt: Agreed, let's remove the initial blank lines from test methods.

foxpuppet/tests/test_browser_model.py
davehunt: Yeah, if tests need this extra level of explanation, let's make it a [docstring](https://www.python.org/dev/peps/pep-0257/) such as:

```python
def test_new_private_window(selenium):
    """Tests opening a new private browsing window via menu"""
```

foxpuppet/browser/window.py
```diff
@@ -0,0 +1,96 @@
...
+
+from .tabbar import TabBar
+from .navbar import Navbar
+from foxpuppet.browser.windows import Windows
+
+NAV_BAR = 'nav-bar'
+TAB_BROWSER = 'tabbrowser-tabs'
+TITLE_BAR_CLOSE = 'titlebar-close'
+TITLE_BAR_MIN = 'titlebar-min'
+TITLE_BAR_MAX = 'titlebar-max'
```
davehunt: These are all IDs, but not all elements will be locatable by ID. I suggest combining the locator with the strategy as a tuple and importing `By` from Selenium. For example:

```python
_nav_bar_locator = (By.ID, 'nav-bar')
```

You can then use `selenium.find_element` like so:

```python
self.selenium.find_element(*self._nav_bar_locator)
```

foxpuppet/browser/base_window.py
```diff
@@ -2,12 +2,8 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from foxpuppet.foxpuppet import FoxPuppet
 
+class BaseWindow(object):
```
davehunt: Is this used? It doesn't appear to do much.

foxpuppet/browser/navbar.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+MENU_BUTTON = 'PanelUI-menu-button'
+
+
+class Navbar(object):
+
+    def __init__(self, selenium, *args, **kwargs):
+        self.selenium = selenium
+        self._locationbar = None
```
davehunt: I think this can be removed?

foxpuppet/tests/test_browser_model.py
whimboo: With stronger rules set for pep8 the docstring might even have to look like "Test opening a new private browsing window via menu.".

foxpuppet/browser/navbar.py
```diff
@@ -0,0 +1,41 @@
...
+        self.wait = WebDriverWait(selenium, 10)
```
davehunt: Let's not worry about it for the initial patch, but I think we should take care of creating the wait and assigning `self.selenium` in a parent class.

foxpuppet/browser/navbar.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def _open_window(self, private=False):
```
davehunt: Why is this method marked as private? I think we can remove the leading `_`.

foxpuppet/browser/navbar.py
```diff
@@ -0,0 +1,41 @@
...
+
+        if private:
+            menu = self.selenium.find_element_by_id(MENU_BUTTON)
+            menu.click()
+            button = self.wait.until(EC.element_to_be_clickable(
+                (By.CSS_SELECTOR, '#privatebrowsing-button')
```
davehunt: All element locators should be defined at the class level.

foxpuppet/browser/navbar.py
```diff
@@ -0,0 +1,41 @@
...
+    def __init__(self, selenium, *args, **kwargs):
+        self.selenium = selenium
+        self._locationbar = None
+        self.wait = WebDriverWait(selenium, 10)
+
+    def _open_window(self, private=False):
+
+        if private:
+            menu = self.selenium.find_element_by_id(MENU_BUTTON)
+            menu.click()
```
davehunt: If you're not using `menu` for anything else you can just to `self.selenium.find_element...(...).click()`

foxpuppet/browser/navbar.py
```diff
@@ -0,0 +1,41 @@
...
+                (By.CSS_SELECTOR, '#privatebrowsing-button')
+            ))
+
+        menu = self.selenium.find_element_by_id(MENU_BUTTON)
+        menu.click()
+        button = self.wait.until(EC.element_to_be_clickable(
+            (By.CSS_SELECTOR, '#new-window-button')
+        ))
+
+        return button
```
davehunt: I don't think this method needs to return anything.

foxpuppet/browser/navbar.py
```diff
@@ -0,0 +1,41 @@
...
+
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support.ui import WebDriverWait
+from selenium.webdriver.support import expected_conditions as EC
+
+
+MENU_BUTTON = 'PanelUI-menu-button'
+
+
+class Navbar(object):
```
davehunt: Nit: TabBar has an uppercase B, we should be consistent.

foxpuppet/browser/tabbar.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+MENU_BUTTON = 'PanelUI-menu-button'
+
+
+class TabBar(object):
+
+    def __init__(self, selenium, *args, **kwargs):
+        self.selenium = selenium
+        self._tabbrowser = None
```
davehunt: Unused? Let's remove it.

foxpuppet/tests/test_browser_model.py
```diff
@@ -0,0 +1,42 @@
...
+def test_new_private_window(selenium):
+
+    foxpuppet = FoxPuppet(selenium)
+    selenium.get('http://www.google.com')
+    selenium.set_context('chrome')
+    foxpuppet.new_window(private=True)
```
davehunt: We need to assert that there are now two windows and one is private.

foxpuppet/tests/test_browser_model.py
```diff
@@ -0,0 +1,42 @@
...
+    assert selenium.title == 'Android'
+
+
+# Tests adding a new bookmark
+def test_bookmark_button(selenium):
+
+    foxpuppet = FoxPuppet(selenium)
+    selenium.set_context('chrome')
+    foxpuppet.bookmark_page()
+
```
davehunt: We should have an assert here

foxpuppet/tests/test_browser_model.py
```diff
@@ -0,0 +1,42 @@
...
+
+def test_new_tab_button(selenium):
+
+    foxpuppet = FoxPuppet(selenium)
+    selenium.get('http://www.google.com')
```
davehunt: No need to load an external website.

foxpuppet/tests/test_browser_model.py
```diff
@@ -0,0 +1,42 @@
...
+    selenium.set_context('chrome')
+    foxpuppet.new_tab()
```
davehunt: Assert the number of tabs.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 selenium == 3.0.0b2
 flake8
 pytest == 2.9.2
+pytest-selenium
```
davehunt: This is a test requirement and should go in `tox.ini`

requirements.txt
```diff
@@ -1,3 +1,4 @@
 selenium == 3.0.0b2
 flake8
 pytest == 2.9.2
```
davehunt: This is a test requirement and should go in `tox.ini`

requirements.txt
```diff
@@ -1,3 +1,4 @@
 selenium == 3.0.0b2
 flake8
```
davehunt: This is not a requirement for the project.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 selenium == 3.0.0b2
```
davehunt: We should use `setup.py` to define the dependencies.

.travis.yml
```diff
@@ -1,18 +1,29 @@
...
-  - "virtualenv venv"
-  - "source venv/bin/activate"
-
-install:
-  - "pip install -r requirements.txt "
+  - pip install mozdownload mozinstall
+  - mozdownload --type daily --branch mozilla-central --destination firefox.tar.bz2
+  - mozinstall firefox.tar.bz2
+  - sudo ln -sf $PWD/firefox/firefox /usr/local/bin/firefox
+  - pip install tox
```
davehunt: Let's instead use [tox-travis](https://pypi.python.org/pypi/tox-travis), which has better support for multiple python versions on Travis.

.travis.yml
```diff
@@ -1,18 +1,29 @@
...
+    - "2.7"
+    - "3.5"
 
-sudo: "required"
+env:
+  global:
+    - DISPLAY=:99.0
+  matrix:
+    - PYTHON_2=2.7
+    - PYTHON_3=3.5
```
davehunt: I think switching to tox-travis (see other comment) would mean we can remove this matrix.

foxpuppet/browser/window.py
```diff
@@ -0,0 +1,96 @@
...
+        self._navbar = None
+        self._tabbar = None
+
+    @property
+    def navbar(self):
+        self.selenium.switch_to.window(self.windows.current)
+
+        if not self._navbar:
+            navbar = self.selenium.find_element_by_id(NAV_BAR)
+            self._navbar = Navbar(self.selenium, navbar)
```
whimboo: That was just a decision to improve performance. The side-effect of caching would be that changes to the underlying DOM node are not getting recognized. Maybe we should better live without that caching nowadays?

foxpuppet/foxpuppet.py
```diff
@@ -2,8 +2,38 @@
...
+        if private:
+            self.window._new_private_browsing_window()
+        else:
+            self.window._new_window()
+
+    def bookmark_page(self):
+        self.window._bookmark_page()
+
+    def new_tab(self):
+        self.window._new_tab()
```
whimboo: Any kind of method which represents an action should get a verb as prefix. `new_tab`is not understandable, but `open_tab()` is much more.

whimboo:
> It's difficult to review this independently from the initial pull request it contains the tests that are under review there. Can we just introduce Tox to the other pull request, and focus on just the one?

I would totally love that too! Please do not mix up different pull requests with the same content to be added. This is really confusing and makes it way harder to keep track of changes. Also it will cause you a lot of burden when it comes to a rebase later for this PR.

whimboo:
Hm, maybe @jrbenny35 accidentally created this new branch based off the branch used for the other pull but not on master? This could actually be the reason for the pollution here.

foxpuppet/browser/window.py
davehunt: Let's remove the caching and revisit if we feel there are worthwhile performance gains.

foxpuppet/browser/base_window.py
jrbenny35: Right now it's not. I think it was going to be the browserwindow, before we changed the hierarchy.

foxpuppet/browser/navbar.py
jrbenny35: This is a big problem. The reason I am doing this is trying to fix the flakiness that happens in travis. This doesn't work however as I get a timeout exception. I am still having problems with travis when it needs to click a button it seems. It will say that the button I want within the menu button, isn't there.

foxpuppet/browser/window.py
```diff
@@ -0,0 +1,96 @@
...
+        windows_to_keep = exceptions or []
+
+        for handle in self.windows.all:
+            if windows_to_keep == handle:
+                continue
+            self.selenium.switch_to.window(handle)
+            self.close(handle)
+
+    def _min_window_size(self):
+        self.selenium.switch_to.window(self.windows.current)
```
jrbenny35: These switch_to calls are kind of placeholders for when we figure out a window handle tracking system. But they could be added later.

jrbenny35:
@whimboo Yeah this branch was originally just to get a travis config running without polluting the first patch. No worries, I will make sure to take all of this into consideration before we close this, and close the other one too.

jrbenny35:
It was really my fault. Trying to get the tests to pass I ended up changing things around.

foxpuppet/browser/navbar.py
whimboo: We haven't used the menu button in our tests yet but I can assume it happens because buttons are getting populated slower than usual on some Travis nodes.

foxpuppet/browser/window.py
whimboo: There is no reason to re-invent things. Use the handles as provided by Selenium which already take care of that.

whimboo: private or not is not a unique identifier of a window! Only handles are and should therefore being used.

foxpuppet/browser/navbar.py
whimboo: In Puppeteer it is afaik. You can check that yourself. 

I even wonder why we don't use as much as possible code from Puppeteer.

foxpuppet/browser/navbar.py
jrbenny35: That seems to be it. 10 seconds isn't enough. I've never tested browser chrome in travis. How should I go about this general slowness?

whimboo: As discussed on the other PR we should rip of this code for now given that we originally only wanted to get the basics of the browser window implemented. This is already something very specific.

foxpuppet/browser/navbar.py
```diff
@@ -0,0 +1,36 @@
...
+    def __init__(self, selenium, *args, **kwargs):
+        self.selenium = selenium
+        self._locationbar = None
+        self._menu_button = (By.ID, 'PanelUI-menu-button')
+
+    def open_window(self, private=False):
+        self.selenium.set_context('chrome')
+        if private:
+            self.selenium.find_element(*self._menu_button).click()
```
whimboo: I really really propose that you make use of the main menu to open a new private browsing window by default. The menu entry is just an ID and is always going to work. This path as used here can fail in various ways.

jrbenny35: You mean going through File > New Private window?

whimboo: Yes, the reason also is that this path is always available. The menu button could have been customized away and reachable via another location. And shortcuts are locale related, so we miss the l10n support. The menu path is also the default for puppeteer.

requirements.txt
whimboo: So FoxPuppet is a library, right? As what I have read in the past for those it's better to define dependencies in requirements.txt. Only end-user applications should define deps in setup.py and then most likely also in a strict way.

davehunt: FoxPuppet will be pip installable from PyPI, not just a library. Otherwise users would have to clone the repository in order to use it.

silne30:
I made some changes by checking out this PR and making some edits. For some reason, I don't see them here. You can check them out on my branch with the same name.

silne30:
OK. Pushed to Benny's fork instead. I didn't get to get the private browsing test completely finished. But the flakiness has been removed. I changed as many locators as I could find. This was a couple hours of work. On Tuesday, I will add some more if I have the bandwidth.

foxpuppet/browser/navbar.py
```diff
@@ -4,33 +4,30 @@
...
 from selenium.webdriver.common.by import By
-from selenium.webdriver.support.ui import WebDriverWait
-from selenium.webdriver.support import expected_conditions as EC
 
 
 class Navbar(object):
+
+    _file_menu_button_locator = (By.ID, 'file-menu')
+    _file_menu_private_window_locator = (By.ID, 'menu_newPrivateWindow')
+    _file_menu_new_window_button_locator = (By.ID, 'menu_newNavigator')
```
whimboo: Those elements are not part of the navbar but the browser window itself.

foxpuppet/browser/navbar.py
```diff
@@ -4,33 +4,30 @@
...
+    _bookmark_menu_button_locator = (By.ID, 'bookmarks-menu-button')
+
     def __init__(self, selenium, *args, **kwargs):
         self.selenium = selenium
         self._locationbar = None
-        self._menu_button = (By.ID, 'PanelUI-menu-button')
 
     def open_window(self, private=False):
```
whimboo: Since when is this method part of the NavBar? Did I miss a move? From my perspective it is still a method which should be attached to the browser window.

foxpuppet/browser/windows.py
```diff
@@ -2,6 +2,8 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
+from selenium.webdriver.common.by import By
+
 
 class Windows(object):
```
whimboo: This looks to be a general windows class not related to the browser only. So it should be kept outside of the browser folder.

jrbenny35:
@whimboo I agree. This one got out of hand because travis wasn't as easy to setup as we thought. If you and @davehunt can hold off on any more updates then we can get this merged sooner than later. We do appreciate all the great feedback! 😄

whimboo:
Maybe its better to get all the module changes into your original PR, close this one, and open a new PR only for teh travis changes. If something fails in the tests we should get those temporarily disabled.

davehunt:
Another suggestion might be to start a new branch from master, and create a very minimal patch that adds a single test and runs in Travis CI. Then we can incrementally add the other parts of this and the other patch. I'd rather not merge stuff just to immediately fix it up.

jrbenny35:
If it's possible can I just finish the reviews on this and merge it? If I make a new patch, all the work that has gotten the tests close to stable will have to be redone.

davehunt:
> If it's possible can I just finish the reviews on this and merge it?

Sounds good to me, though let's try not to let the scope creep.

> If I make a new patch, all the work that has gotten the tests close to stable will have to be redone.

No work would need to be redone, as you'd just be able to copy/paste the essential parts, but I'm also fine with finishing up this patch. Can you close whichever pull request we're not going to be working on then?

foxpuppet/tests/test_browser_model.py
```diff
@@ -0,0 +1,39 @@
...
+
+from foxpuppet import FoxPuppet
+
+
+class TestBrowserModel(object):
+
+    def test_new_private_window(self, selenium):
+        """Tests opening a new private browsing window via menu"""
+        foxpuppet = FoxPuppet(selenium)
+        with foxpuppet.windows.wait_for_new_window():
```
whimboo: This should really be done transparently. If you have to to do it each time when you open a window, it will blow up the tests horribly. Why don't you pick the code from puppeteer's windows.py module?

jrbenny35: Selenium doesn't have context manager integrated yet. John's patch just landed so we are waiting for a new selenium release.

silne30: @whimboo I agree.

@jrbenny35 Thanks for the note. What Henrik means is that you shouldn't have to do this in every test. This is a new context manager I wrote. Not to manage the scope of the test but to manage the wait for the window to open. It does need to be cleaned up so that the test writer does not have to worry about adding the wait.

jrbenny35: @silne30 Ahh, okay. Dave had said something about context manager merged into Selenium. I mixed them up. I will add them in the appropriate places though.

jrbenny35:
@davehunt I think we have most of this done for the initial merge. We will have to wait for selenium to merge Johns patch for the context switching but that is an easy change. Please look these recent changes over when you can!

.travis.yml
davehunt: You're installing tox-travis on line 26, but you should install it here instead of tox.

.travis.yml
```diff
@@ -1,18 +1,29 @@
...
+  - pip install tox
+  - wget -O geckodriver.tar.gz https://github.com/mozilla/geckodriver/releases/download/v0.11.0/geckodriver-v0.11.0-linux64.tar.gz
+  - gunzip -c geckodriver.tar.gz | tar xopf -
+  - chmod +x geckodriver && sudo mv geckodriver /usr/local/bin
+  - firefox --version
+  - geckodriver --version
 
 install:
-  - "pip install -r requirements.txt "
+  pip install tox-travis
```
davehunt: You don't really need an install section, as tox will take care of that.

.travis.yml
```diff
@@ -1,18 +1,29 @@
...
 before_install:
-  - "pip install virtualenv"
-  - "virtualenv venv"
-  - "source venv/bin/activate"
+  - pip install mozdownload mozinstall
+  - mozdownload --type daily --branch mozilla-central --destination firefox.tar.bz2
+  - mozinstall firefox.tar.bz2
+  - sudo ln -sf $PWD/firefox/firefox /usr/local/bin/firefox
+  - pip install tox
+  - wget -O geckodriver.tar.gz https://github.com/mozilla/geckodriver/releases/download/v0.11.0/geckodriver-v0.11.0-linux64.tar.gz
```
davehunt: Latest is 0.11.1

.travis.yml
```diff
@@ -1,18 +1,29 @@
...
-sudo: "required"
+env:
+  global:
+    - DISPLAY=:99.0
 
 before_install:
-  - "pip install virtualenv"
-  - "virtualenv venv"
-  - "source venv/bin/activate"
+  - pip install mozdownload mozinstall
```
davehunt: These won't work with python 3.5 so either remove python 3.5 (and re-introduce it in another patch), or use a different method for downloading/installing Firefox.

.travis.yml
```diff
@@ -21,7 +32,7 @@ notifications:
     on_start: "change"
 
 before_script:
-  - "flake8 ."
+  - sh -e /etc/init.d/xvfb start
 
 script:
-  - "py.test --capture=sys --showlocals --tb=long foxpuppet/tests "
\ No newline at end of file
+  - tox
```
davehunt: You should call tox-travis here

foxpuppet/browser/base_window.py
davehunt: This still isn't used yet, let's remove it and reintroduce it when it's needed.

foxpuppet/browser/navbar.py
```diff
@@ -0,0 +1,18 @@
...
+class Navbar(object):
+    _bookmark_menu_button_locator = (By.ID, 'bookmarks-menu-button')
+
+    def __init__(self, selenium, *args, **kwargs):
+        self.selenium = selenium
+
+    def bookmark_page(self):
+        self.selenium.set_context('chrome')
+        self.selenium.find_element(
+            *self._bookmark_menu_button_locator).click()
```
davehunt: It concerns my that there is no wait here, meaning we return before the page has been successfully bookmarked.

foxpuppet/browser/tabbar.py
```diff
@@ -0,0 +1,17 @@
...
+
+class Tabbar(object):
+    _new_tab_locator = (By.ID, 'new-tab-button')
+
+    def __init__(self, selenium, *args, **kwargs):
+        self.selenium = selenium
+
+    def open_tab(self):
+        self.selenium.set_context('chrome')
+        self.selenium.find_element(*self._new_tab_locator).click()
```
davehunt: Again, we should have a wait that means we only return once the new tab is open.

foxpuppet/browser/window.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def __init__(self, selenium, *args, **kwargs):
+        self.selenium = selenium
+        self.windows = Windows(selenium)
+        self.navbar = Navbar(selenium)
+        self.tabbar = Tabbar(selenium)
+
+    @property
+    def _navbar(self):
+        self.selenium.switch_to.window(self.windows.current)
```
davehunt: This is a noop - if the window is current then we don't need to switch to it.

foxpuppet/browser/window.py
```diff
@@ -0,0 +1,80 @@
...
+
+        if not self._navbar:
+            navbar = self.selenium.find_element(*self._nav_bar_locator)
+            return Navbar(self.selenium, navbar)
+        return self._navbar
```
davehunt: Didn't we decide to remove the caching, and revisit it later if we felt there were performance issues?

foxpuppet/browser/window.py
```diff
@@ -0,0 +1,80 @@
...
+    def is_private(self):
+        """Returns True if this is a Private Browsing window."""
+
+        self.selenium.set_context('chrome')
+        return self.selenium.execute_script("""
+                Components.utils.import("resource://gre/modules/PrivateBrowsingUtils.jsm");
+
+                let chromeWindow = arguments[0].ownerDocument.defaultView;
+                return PrivateBrowsingUtils.isWindowPrivate(chromeWindow);
+            """, self.windows.window_element)
```
davehunt: It feels like a `BrowserWindow` object should have a `_root_element` property that returned the root element, rather than calling `self.windows`, which from the naming suggests it represents a collection of windows.

foxpuppet/browser/window.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def open_window(self, private=False):
+        self.selenium.set_context('chrome')
+        self.selenium.find_element(*self._file_menu_button_locator).click()
+        with self.windows.wait_for_new_window():
```
davehunt: Interesting use of a context manager.

foxpuppet/browser/window.py
```diff
@@ -0,0 +1,80 @@
...
+            if private:
+                self.selenium.find_element(
+                    *self._file_menu_private_window_locator).click()
+            else:
+                self.selenium.find_element(
+                    *self._file_menu_new_window_button_locator).click()
+        self.selenium.set_context('content')
```
davehunt: I think it would be valuable to return the handle for the new window here.

foxpuppet/browser/window.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def close(self):
+        self.selenium.close()
+
+    def minimize(self):
+        # TODO: Change minimize to work from menu
```
davehunt: Is this TODO captured elsewhere?

foxpuppet/browser/window.py
```diff
@@ -0,0 +1,80 @@
...
+        button = self.selenium.find_element(
+            *self._title_bar_minimize_button_locator)
+        button.click()
```
davehunt: Again, we probably want a wait to make sure the window has finished minimizing.

foxpuppet/browser/window.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def maximize(self):
+        self.selenium.maximize()
```
davehunt: This isn't using Firefox UI, which is the point of FoxPuppet - could you add options for maximizing via the UI?

foxpuppet/tests/test_browser_model.py
```diff
@@ -0,0 +1,37 @@
...
+        foxpuppet.browser.open_window(private=False)
+        assert len(foxpuppet.browser.windows.all) == 2
+        assert foxpuppet.browser.is_private is False
+        foxpuppet.browser.windows.focus(foxpuppet.browser.windows.all[1])
+        assert foxpuppet.browser.is_private is False
+
+    def test_bookmark_button(self, selenium):
+        """"Tests adding a new bookmark"""
+        foxpuppet = FoxPuppet(selenium)
+        foxpuppet.browser.navbar.bookmark_page()
```
davehunt: Need an assert here.

foxpuppet/tests/test_browser_model.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def test_new_tab_button(self, selenium):
+        """Tests opening a new tab via tab bar"""
+        foxpuppet = FoxPuppet(selenium)
+        foxpuppet.browser.tabbar.open_tab()
```
davehunt: Need an assert here.

foxpuppet/windows.py
```diff
@@ -0,0 +1,44 @@
...
+            lambda driver: len(handles_before) != len(driver.window_handles))
+
+    @property
+    def window_element(self):
+        """Returns the inner DOM window element.
+
+        :returns: DOM window element.
+        """
+
+        self.selenium.switch_to.window
```
davehunt: This isn't doing anything.

requirements.txt
```diff
@@ -1,3 +1,2 @@
-selenium == 3.0.0b2
-flake8
+selenium == 3.0.1
 pytest == 2.9.2
```
davehunt: Remove this file. Dependencies are specified in `setup.py` and `tox.ini`

setup.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+from setuptools import setup, find_packages
+
+
+setup(
+    name='FoxPuppet',
+    version='0.1.0',
+    description='Foxpuppet for Firefox UI testing',
+    long_description='See http://foxpuppet.readthedocs.org/',
```
davehunt: We should add a README and use that for the `long_description`. See https://github.com/pytest-dev/pytest-selenium/blob/master/setup.py#L6 for an example.

setup.py
```diff
@@ -0,0 +1,22 @@
...
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+
+from setuptools import setup, find_packages
+
+
+setup(
+    name='FoxPuppet',
+    version='0.1.0',
+    description='Foxpuppet for Firefox UI testing',
```
davehunt: I think we need a better description here. Something like 'Firefox user interface model for use with Selenium'. It's hard to describe, but I think we should mention Selenium and I like to avoid indicating that it's only for testing. I think there are many creative things users could do that aren't considered testing.

setup.py
```diff
@@ -0,0 +1,22 @@
...
+    long_description='See http://foxpuppet.readthedocs.org/',
+    url='https://github.com/mozilla/FoxPuppet',
+    license='MPL',
```
davehunt: MPL2. See https://github.com/pytest-dev/pytest-selenium/blob/master/setup.py#L31

setup.py
```diff
@@ -0,0 +1,22 @@
...
+    packages=find_packages(),
```
davehunt: I prefer to list the packages, for more control.

setup.py
```diff
@@ -0,0 +1,22 @@
...
+    include_package_data=True,
```
davehunt: Which package data? I think we can remove this.

setup.py
```diff
@@ -0,0 +1,22 @@
...
+    install_requires=[
+        'pytest',
+        'selenium'
```
davehunt: We should specify a minimum version of 3.0.1 (current), and we'll need to bump this when the 'context' context manager is in a released version of the client.

setup.py
```diff
@@ -0,0 +1,22 @@
...
+    name='FoxPuppet',
+    version='0.1.0',
+    description='Foxpuppet for Firefox UI testing',
+    long_description='See http://foxpuppet.readthedocs.org/',
+    url='https://github.com/mozilla/FoxPuppet',
+    license='MPL',
+    packages=find_packages(),
+    include_package_data=True,
+    install_requires=[
+        'pytest',
```
davehunt: FoxPuppet does not require pytest.

tox.ini
```diff
@@ -0,0 +1,18 @@
+[tox]
+skipsdist = true
```
davehunt: We don't want to skip creating the dist.

tox.ini
```diff
@@ -0,0 +1,18 @@
...
+envlist = py{27,35}
+
+[tox:travis]
+2.7 = py27
+3.4 = py34
```
davehunt: We're using 3.5 in Travis, but anyway I think this section is not needed as tox-travis should already be able to map these.

tox.ini
```diff
@@ -0,0 +1,18 @@
...
+
+[testenv]
+passenv = DISPLAY
+deps =
+    pytest==3.0.3
+    pytest-selenium==1.4.0
```
davehunt: Please upgrade to 1.5.0

tox.ini
```diff
@@ -0,0 +1,18 @@
...
+    selenium==3.0.1
+commands = pytest --driver Firefox {posargs}
+
+[pytest]
+testpaths = foxpuppet/tests
```
davehunt: Tests should not be included in the `foxpuppet` package - please move them to the top level.

.travis.yml
```diff
@@ -1,18 +1,29 @@
...
 python:
-  - "2.7"
-  - "3.5"
+    - "2.7"
+    - "3.5"
 
-sudo: "required"
+env:
+  global:
+    - DISPLAY=:99.0
```
davehunt: Add a matrix using `TOXENV` so we have one job running `py27` and another using `tox`:

```yml
env:
  global:
    - DISPLAY=:99.0
  matrix:
    - TOXENV=flake8
matrix:
  include:
    - python: 2.7
      env: TOXENV=py27
    - python: 3.5
      env: TOXENV=py35
```

You might need to play with this a little, and there could be a better way to get all the jobs to run. Basically you want the following jobs:

 * Python 2.7
    * flake8
    * py27
 * Python 3.5
    * flake8
    * py35

We should also add building the docs to Travis so we get a notification if a pull request brakes these. If it makes it easier, drop python 3.5 from this patch and we'll work on that next.

foxpuppet/browser/window.py
jrbenny35: There is a way to do it on osx, but linux is different.

.travis.yml
jrbenny35: I am working on a patch for travis. I'd like to get python 3.5 in.

foxpuppet/browser/window.py
whimboo: So you consider it better than a method like open_window() which implicitly waits?

foxpuppet/browser/window.py
whimboo: Correct.

foxpuppet/browser/tabbar.py
whimboo: We should also use `using_context` to not confuse test authors with context switches in the test.

whimboo:
> I think our window API needs some more thought. At the moment you can access the initial browser window through foxpuppet.browser, but this also has a windows property. As you can have multiple browser windows, I think it might make more sense to use foxpuppet.windows as a collection of all windows that can be used to filter by type. We could still have a foxpuppet.browser helper, which returns the initial browser window. Do you have some thoughts on this @whimboo?

This is exactly what Puppeteer has implemented. I still don't understand why we completely re-inventing all the API for foxpuppet again. So far I haven't gotten a reply to that question, which I raise I think the third time now.

jrbenny35:
@whimboo It was changed, I just don't think john saw it when he re-wrote the tests. If you look in foxpuppet.py the windows object is assigned, however in window.py, windows was assigned again. It was changed, I just forgot to remove it from windows.py

.travis.yml
```diff
@@ -15,8 +15,7 @@ before_install:
   - mozdownload --type daily --branch mozilla-central --destination firefox.tar.bz2
   - mozinstall firefox.tar.bz2
   - sudo ln -sf $PWD/firefox/firefox /usr/local/bin/firefox
-  - pip install tox
-  - wget -O geckodriver.tar.gz https://github.com/mozilla/geckodriver/releases/download/v0.11.0/geckodriver-v0.11.0-linux64.tar.gz
+  - wget -O geckodriver.tar.gz https://github.com/mozilla/geckodriver/releases/download/v0.11.1/geckodriver-v0.11.0-linux64.tar.gz
```
davehunt: The version number appears twice in this line, so you need to update them both.

tests/test_browser_model.py
```diff
@@ -12,18 +12,18 @@ def test_new_private_window(self, selenium):
         """Tests opening a new private browsing window via menu"""
         foxpuppet = FoxPuppet(selenium)
         foxpuppet.browser.open_window(private=True)
-        assert len(foxpuppet.browser.windows.all) == 2
+        assert len(foxpuppet.windows.all) == 2
         assert foxpuppet.browser.is_private is False
-        foxpuppet.browser.windows.focus(foxpuppet.browser.windows.all[1])
+        foxpuppet.windows.focus(foxpuppet.windows.all[1])
```
davehunt: I would say that `foxpuppet.windows` should return a collection of window objects. This should therefore be `foxpuppet.windows[1].focus` and line 15 should be `len(foxpuppet.windows)`.

tests/test_new_window_handle.py
```diff
@@ -0,0 +1,17 @@
+from foxpuppet import FoxPuppet
+from time import sleep
+
+
+class Test_new_window_handle:
+
+    def test_new_private_window(self, selenium):
+
+        foxpuppet = FoxPuppet(selenium)
+        selenium.get('http://www.google.com')
```
davehunt: Please stop writing tests that depend on Google or any other external website. 😄

tests/test_new_window_handle.py
```diff
@@ -0,0 +1,17 @@
...
+        selenium.set_context('chrome')
+        menu = selenium.find_element_by_id('main-menubar')
+        menu.get_attribute('File')
+        menu.click()
+        menu.get_attribute('New Private Window')
+        menu.click()
+        sleep(5)
```
davehunt: This isn't using FoxPuppet, but Selenium.. also, `sleep` is good for you, but bad for your tests. 😴

foxpuppet/browser/tabbar.py
davehunt: @whimboo the 'context' context manager is not yet available in the Selenium python client. The code has been merged and it will be in the next release.

foxpuppet/browser/window.py
whimboo: This is more a question about implicit vs. explicit waiting for a window
to have opened and finished loading. I actually do not see a reason why
we have to force each test to explicitly check that the window has been
opened while all that can be done inside of open_window().

davehunt: This isn't a test, it's the `open_window` method that the test would call. This is therefore an implicit wait from the perspective of the user, but an explicit wait by implementation. Note that 'implicit waits' is a term often reserved in Selenium for element lookup on the server side. We should not include calls to wait for new windows in the tests, but I don't see that happening here.

whimboo: Oh! I totally overlooked that. Right, this is in the module. So a question which raises for me now is if there is a case when multiple windows are getting opened, how could we specify that with the context handler. This will be very rare but just to bring it up.

tests/test_new_window_handle.py
jrbenny35: This test wasn't supposed to be included.

tests/test_browser_model.py
jrbenny35: So what would be the purpose of focus if we are already going to switch to a specific window from the list? Maybe we don't need focus because we can just use the switch_to.window() in selenium.

foxpuppet/browser/window.py
jrbenny35: Dave do we even need to minimize and maximize windows? I can try to use the menu > windows > minimize for the minimize method, but there isn't anything for maximize that I can see.

foxpuppet/browser/window.py
jrbenny35: What do you mean by captured?

whimboo:
This is more a question about implicit vs. explicit waiting for a window
to have opened and finished loading. I actually do not see a reason why
we have to force each test to explicitly check that the window has been
opened while all that can be done inside of open_window().

foxpuppet/browser/window.py
whimboo: There is "Zoom" in the same menu which seems to be what you want for maximize.

foxpuppet/browser/window.py
davehunt: I mean open an issue in GitHub or similar.

foxpuppet/browser/window.py
davehunt: You raise a good point - do we need to allow users to minimize the browser window? I'm not sure it's particularly useful for testing the UI... @whimboo do you need this for your UI tests? The conversation about maximize doesn't really need to happen on this patch. Why don't we just remove this code for now and open an issue for it?

jrbenny35:
I got the flake8 job running but when we add python 3.5 it will need to be changed. Everything should be up to date with the requested changes.

jrbenny35:
What kind of documentation are we doing? Maybe I can start on that

jrbenny35:
So travis may not be able to help us with getting the nightlys. We may be able to get dev editions though. I think, for now, I could write a python script that would download and install the nightlys that we need. Until we can get something through on travis.

whimboo:
If you really want to use foxpuppet for different versions of Firefox we strongly have to talk about backward compatibility and how we want to release foxpuppet for different branches including updates for late ui changes.

With all that I'm not that confident anymore to have the repository separated from the Firefox source tree. From my experience over the last years with Mozmill and Firefox UI tests I can say it's a pain.

davehunt:
> What kind of documentation are we doing? Maybe I can start on that

Let's use readthedocs. I've created a project and added @whimboo as a maintainer. The docs will be available at http://foxpuppet.readthedocs.io/en/latest/ once there's something ready to build. We already have #23 #24 and #25, so let's pick one of those (maybe close the others?) and move conversation of the docs to whichever one we pick.

davehunt:
> If you really want to use foxpuppet for different versions of Firefox we strongly have to talk about backward compatibility and how we want to release foxpuppet for different branches including updates for late ui changes.

> With all that I'm not that confident anymore to have the repository separated from the Firefox source tree. From my experience over the last years with Mozmill and Firefox UI tests I can say it's a pain.

I totally agree, which is why we've decided to just support one version initially (other versions may work but will not be guaranteed). We've already acknowledged that we expect the source code to move into mozilla-central ultimately, but want to get it started in GitHub so we can develop rapidly. If we have multiple versions tested in Travis CI then we can mark some as allowed to fail.

docs/Makefile
```diff
@@ -0,0 +1,225 @@
+# Makefile for Sphinx documentation
```
davehunt: Let's move the docs to a separate pull request.

foxpuppet/browser/base_window.py
davehunt: As mentioned previously, let's remove this unused code for now.

foxpuppet/browser/navbar.py
davehunt: We discussed reducing this patch to open new window, new private browsing window, and new tab. Let's remove the bookmark components for now.

foxpuppet/browser/tabbar.py
davehunt: There's still no wait here. If this is a challenge then let's drop open tab from this patch.

foxpuppet/browser/window.py
```diff
@@ -0,0 +1,50 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from foxpuppet.windows import Windows
+from selenium.webdriver.common.by import By
```
davehunt: From [PEP8](https://www.python.org/dev/peps/pep-0008/#imports):

>  Imports should be grouped in the following order:
>
> 1.   standard library imports
> 1.   related third party imports
> 1.   local application/library specific imports
>
>You should put a blank line between each group of imports.

foxpuppet/browser/window.py
```diff
@@ -0,0 +1,50 @@
...
+
+from .navbar import Navbar
```
davehunt: If we remove the bookmark parts we can remove the references to navbar.

foxpuppet/browser/window.py
```diff
@@ -0,0 +1,50 @@
...
+from .tabbar import Tabbar
```
davehunt: If we decide to remove open tab from this patch we can also remove the references to toolbar.

foxpuppet/browser/window.py
```diff
@@ -0,0 +1,50 @@
...
+    _file_menu_private_window_locator = (By.ID, 'menu_newPrivateWindow')
+    _file_menu_new_window_button_locator = (By.ID, 'menu_newNavigator')
+    _nav_bar_locator = (By.ID, 'nav-bar')
+    _tab_browser_locator = (By.ID, 'tabbrowser-tabs')
+
+    def __init__(self, selenium, *args, **kwargs):
+        self.selenium = selenium
+        self.navbar = Navbar(selenium)
+        self.tabbar = Tabbar(selenium)
+        self._windows = Windows(selenium)
```
davehunt: Just a note that we're going to follow up the window manager in the next pull.

foxpuppet/windows.py
```diff
@@ -0,0 +1,48 @@
...
+        return self.selenium.current_window_handle
+
+    def focus(self, handle):
+        return self.selenium.switch_to.window(handle)
+
+    def get_new_window_handle(self, handles):
+        """Get new window handle"""
+        if handles != self.selenium.window_handles:
+            new_handle = self.selenium.window_handles[len(handles):]
+        return new_handle
```
davehunt: This is making too many assumptions. If there are less handles then the last handle will be returned, if there are multiple new handles then only the last one will be returned. Ideally the code that launches the new window and waits for the new handle to appear should return the handle.

foxpuppet/windows.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @contextmanager
+    def wait_for_new_window(self):
+        handles_before = self.selenium.window_handles
+        yield
+        WebDriverWait(self.selenium, 10).until(
+            lambda driver: len(handles_before) != len(driver.window_handles))
```
davehunt: I wonder if it would make more sense to register an event listener before we trigger the window open, and perhaps the listener would have a reference to the window handle, which we could use? Do you have any suggestions @whimboo?

Looking at Puppeteer, there's some code we could reuse [here](https://dxr.mozilla.org/mozilla-central/source/testing/puppeteer/firefox/firefox_puppeteer/ui/windows.py#343-381). It looks like we're blocked on [bug 1121698](https://bugzilla.mozilla.org/show_bug.cgi?id=1121698) for making this more robust.

tox.ini
```diff
@@ -0,0 +1,18 @@
+[tox]
+envlist = py27, flake8
+
+[testenv]
+passenv = DISPLAY
+deps =
+    pytest==3.0.3
+    pytest-selenium==1.5.0
+    selenium==3.0.1
```
davehunt: FoxPuppet already depends on selenium, so we can remove this.

tox.ini
```diff
@@ -0,0 +1,18 @@
...
+commands = pytest --driver Firefox {posargs}
+
+[testenv:flake8]
+deps =
+	flake8
```
davehunt: Keep this all on one line please.

tox.ini
```diff
@@ -0,0 +1,18 @@
...
+commands = flake8 foxpuppet tests
```
davehunt: We should run this on the current directory rather than list this. You can make this overridable using `{posargs:.}`

tests/test_browser_model.py
```diff
@@ -0,0 +1,37 @@
...
+
+class TestBrowserModel(object):
+
+    def test_new_private_window(self, selenium):
+        """Tests opening a new private browsing window via menu"""
+        foxpuppet = FoxPuppet(selenium)
+        foxpuppet.browser.open_window(private=True)
+        assert len(foxpuppet.windows.all) == 2
+        assert foxpuppet.browser.is_private is False
+        foxpuppet.windows.focus(foxpuppet.windows.all[1])
```
davehunt: with `open_window` returning the handle (ultimately we might have it return a window object) we can just use this value to switch to it using Selenium.

tests/test_browser_model.py
```diff
@@ -0,0 +1,37 @@
...
+        assert foxpuppet.browser.is_private is True
```
davehunt: `foxpuppet.browser` should still be the main window. Perhaps this is something to clean up with the window manager.

tests/test_browser_model.py
```diff
@@ -0,0 +1,37 @@
...
+        foxpuppet.browser.open_window(private=False)
+        assert len(foxpuppet.windows.all) == 2
+        assert foxpuppet.browser.is_private is False
+        foxpuppet.windows.focus(foxpuppet.windows.all[1])
+        assert foxpuppet.browser.is_private is False
+
+    def test_bookmark_button(self, selenium):
+        """"Tests adding a new bookmark"""
+        foxpuppet = FoxPuppet(selenium)
+        foxpuppet.browser.navbar.bookmark_page()
```
davehunt: No asserts, I recommend removing this test and associated code.

tests/test_browser_model.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def test_new_tab_button(self, selenium):
+        """Tests opening a new tab via tab bar"""
+        foxpuppet = FoxPuppet(selenium)
+        foxpuppet.browser.tabbar.open_tab()
```
davehunt: No asserts, I recommend removing this test and associated code.

tox.ini
```diff
@@ -0,0 +1,18 @@
...
+[testenv]
+passenv = DISPLAY
+deps =
+    pytest==3.0.3
+    pytest-selenium==1.5.0
+    selenium==3.0.1
+commands = pytest --driver Firefox {posargs}
+
+[testenv:flake8]
+deps =
```
davehunt: You can add `skip_install = true` to this environment.

foxpuppet/browser/window.py
jrbenny35: Has this been decided? I don't remember us talking about this. Same as for the bookmark method.

foxpuppet/windows.py
jrbenny35: Well this would only return the newest handle. Which means you would need to know if you need a specific handle when you create it. If this isn't correct, i'll have to adapt the Puppeteer setup.

tests/test_browser_model.py
jrbenny35: What do you mean by main window?

foxpuppet/windows.py
davehunt: You're assuming the window handles are in some sort of order? I wouldn't make that assumption. If you restricted this to ensure that there's only one new handle then it would be much safer, however then a standalone method for getting a new window handle has limited user - it will only ever be called when opening a new window and should probably therefore be marked as private.

tests/test_browser_model.py
davehunt: I mean the initial browser window. This is something we discussed last week with @whimboo and is how Puppeteer works currently.

foxpuppet/windows.py
davehunt: Yes, and I'm not disputing that. How do you know there aren't two 'new' window handles, or what would happen if there were less handles?

jrbenny35:
Is there a way I can get tox to stop all jobs if the flake8 fails? Since they are run on different environments it seems tricky.

davehunt:
>Is there a way I can get tox to stop all jobs if the flake8 fails? Since they are run on different environments it seems tricky.

You can try [fast finishing](https://docs.travis-ci.com/user/customizing-the-build/#Fast-Finishing) but I usually let the tests run because they may be failing for other reasons that are worth investigating independently from flake8 issues.

jrbenny35:
I was also thinking that. Well since we can install firefox from travis i'll add back 3.5 as well.

.travis.yml
```diff
@@ -8,12 +8,16 @@ addons:
...
   global:
     - DISPLAY=:99.0
-  matrix:
-    - TOXENV=flake8
 matrix:
   include:
     - python: 2.7
+      env: TOXENV=flake8
+    - python: 2.7
       env: TOXENV=py27
```
davehunt: This one _shouldn't_ be needed - it's what tox-travis takes care of.

.travis.yml
```diff
@@ -8,12 +8,16 @@ addons:
...
+    - python: 3.5
+      env: TOXENV=flake8
+    - python: 3.5
+      env: TOXENV=py35
```
davehunt: As above, this shouldn't be needed. See https://github.com/pytest-dev/pytest-html/blob/master/.travis.yml for an example of using multiple python versions.

jrbenny35: Got it! Tox travis is nice!

foxpuppet/windows.py
jrbenny35: This would be good for a separate PR.

tests/test_browser_model.py
jrbenny35: So is this fine for now?

jrbenny35:
Hey Dave, think I have caught up to your review. Some changes I don't know if they should be done now or wait until after this patch. I have a good understanding of the window management from puppeteer, so I think that would be the nest PR.

README.rst
```diff
@@ -0,0 +1,15 @@
+FoxPuppet
+=========
+
+Fox Puppet is a library for desktop firefox that provides ui interaction and state using the `Selenium <http://seleniumhq.org/>`_ library.
+
+.. image:: https://img.shields.io/badge/license-MPL%202.0-blue.svg
+   :target: https://github.com/pytest-dev/pytest-selenium/blob/master/LICENSE
```
davehunt: Please update this to point to the FoxPuppet repository.

davehunt: Also, this file does not exist so we'll need to add it in this patch.

README.rst
```diff
@@ -0,0 +1,15 @@
...
+   :alt: License
+
+.. image:: https://img.shields.io/github/issues/mozilla/FoxPuppet.svg
+   :target: https://github.com/mozilla/FoxPuppet/issues
+   :alt: Issues
+   
+Resources
+---------
```
davehunt: If we don't have any resources yet, we should remove the heading.

README.rst
```diff
@@ -4,12 +4,9 @@ FoxPuppet
 Fox Puppet is a library for desktop firefox that provides ui interaction and state using the `Selenium <http://seleniumhq.org/>`_ library.
 
 .. image:: https://img.shields.io/badge/license-MPL%202.0-blue.svg
-   :target: https://github.com/pytest-dev/pytest-selenium/blob/master/LICENSE
+   :target: https://github.com/mozilla/FoxPuppet
```
davehunt: You're missing /blob/master/LICENSE and the LICENSE file itself.

PR #36, opened 2016-11-11
Remove old readme @davehunt

jrbenny35:
This will remove the old readme file.


jrbenny35:
@davehunt I think I got it. Had to figure out why there were so many commits.

PR #37, opened 2016-11-22
Add window management.

jrbenny35:
@davehunt, @whimboo

Just wanted to get this started so you could see what path I am on and I can ask questions in one place.

I have moved some things around and added some of the methods from the window management classes in Puppeteer.


jrbenny35:
@whimboo

Could you please elaborate on this window_type class in marionette?
https://dxr.mozilla.org/mozilla-central/source/testing/puppeteer/firefox/firefox_puppeteer/ui/windows.py#109

Found here:
https://dxr.mozilla.org/mozilla-central/source/testing/marionette/client/marionette_driver/marionette.py#1627

tox.ini
```diff
@@ -13,5 +13,8 @@ skip_install = true
 deps = flake8
 commands = flake8 {posargs:.}
 
+[flake8]
+max-line-length = 99
```
davehunt: I'd personally prefer to use the default here.

tests/test_browser_model.py
```diff
@@ -3,7 +3,7 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
 
-from foxpuppet import FoxPuppet
+from foxpuppet.foxpuppet import FoxPuppet
```
davehunt: We should keep this as `from foxpuppet import FoxPuppet` for the sake of our users sanity. :smile:

foxpuppet/foxpuppet.py
```diff
@@ -2,13 +2,22 @@
...
+
+    @property
+    def selenium(self):
+        return self._selenium
+
+    @use_class_as_property('api.windows.Windows')
+    def windows(self):
+        pass
+
+    @use_class_as_property('api.browser.window.BrowserWindow')
```
davehunt: Why do we need this?

foxpuppet/api/windows.py
```diff
@@ -0,0 +1,173 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
davehunt: I'm against moving everything into an `api` directory. FoxPuppet is entirely an API, so I feel the extra level is redundant. It also makes reviewing very difficult when combined with other changes.

jrbenny35: Everything was moved because I wanted to be able to extend foxpuppet.py in some of the other classes, to have access to it. It felt, to me, that it made sense to not have anything on the same 'level' as foxpuppet.py for OOP sake.

davehunt: Why do you want to extend foxpuppet.py?

jrbenny35: To have access to the selenium object

foxpuppet/foxpuppet.py
davehunt: I don't understand, can you give an example of your circular import issues?

foxpuppet/api/windows.py
davehunt: Can't you pass the selenium object when instantiating your window manager?

jrbenny35: Yeah, that how it was before.

whimboo: First I agree with Dave that we should not move the files under `api/`. For Puppeteer this folder only contains Gecko-API related modules. Foxpuppet is a helper for UI interaction, so it's fine to have all in the top-level folder.

Also please do not remove files and create new ones with the same content. Instead make use of `git mv` to move the files.

foxpuppet/api/windows.py
```diff
@@ -0,0 +1,173 @@
...
+
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support.wait import WebDriverWait
+
+from foxpuppet.foxpuppet import FoxPuppet
+
+
+class Windows(FoxPuppet):
+
+    def __init__(self, selenium):
```
whimboo: Only use selenium here when you have to use it. Otherwise use *args and **kwargs and pass everything through to the super class.

foxpuppet/foxpuppet.py
whimboo: I think we want to drop this decorator completely. It wasn't such a great idea to use in Puppeteer either.

jrbenny35:
Hmm, so seems you both have the same thoughts. Well I will start over then since this PR is pretty indepth. Thanks!

whimboo:
For the future I would suggest to ask earlier for feedback. That will allow us to iterate faster and don't have to waste someones time in adding code which will partly be removed again.

PR #38, opened 2016-11-29
Add more window management 

jrbenny35:
@davehunt @whimboo

So the last attempt was a bit too complex and too much like puppeteer, so I am going to try and simplify and only take what we need. 

I am starting this PR early so I don't get to far off track. 

The tests will probably fail due to me using a custom build of selenium.


foxpuppet/windows.py
```diff
@@ -43,3 +47,28 @@ def window_element(self):
...
+    def switch_to(self, target):
+        """Switches context to the specified chrome window"""
+
+        target_handle = None
+
+        # Check if requested handle is available
+        if target in self.all:
+            target_handle = target
+        else:
+            print("Handle not found")
```
davehunt: I don't think we need this check, if the handle is not available then we should let Selenium raise an exception.

foxpuppet/windows.py
```diff
@@ -43,3 +47,28 @@ def window_element(self):
...
+
+        if not target_handle:
+            self.selenium.switch_to.window(self.current)
```
davehunt: This is a no-op, as you're switching to the window that's already the current window.

foxpuppet/windows.py
```diff
@@ -43,3 +47,28 @@ def window_element(self):
...
+
+        # Switch to window if not currently on it
+        if target_handle != self.current:
+            self.selenium.switch_to.window(target_handle)
```
davehunt: I don't know why a user would try switching to the current window, but it shouldn't cause any problems. I wonder if we're writing unnecessary code here. A user (or FoxPuppet) can just call `self.selenium.switch_to.window(handle)` directly and allow any exceptions to be thrown. I don't feel like we're gaining anything be abstracting this call.

davehunt:
> The tests will probably fail due to me using a custom build of selenium.

I released 3.0.2 earlier this week with the context manager for chrome/content.

foxpuppet/browser/window.py
```diff
@@ -22,23 +22,22 @@ def __init__(self, selenium, *args, **kwargs):
     def is_private(self):
         """Returns True if this is a Private Browsing window."""
 
-        self.selenium.set_context('chrome')
-        return self.selenium.execute_script("""
-                Components.utils.import("resource://gre/modules/PrivateBrowsingUtils.jsm");
+        with self.selenium.context('chrome'):
+            return self.selenium.execute_script("""
+                    Components.utils.import("resource://gre/modules/PrivateBrowsingUtils.jsm");
```
whimboo: 4 chars of indentation please.

foxpuppet/browser/window.py
```diff
@@ -22,23 +22,22 @@ def __init__(self, selenium, *args, **kwargs):
...
+        with self.selenium.context('chrome'):
+            self.selenium.find_element(*self._file_menu_button_locator).click()
+            with self._windows.wait_for_new_window():
+                if private:
+                    self.selenium.find_element(
+                        *self._file_menu_private_window_locator).click()
+                else:
+                    self.selenium.find_element(
+                        *self._file_menu_new_window_button_locator).click()
+        return self.selenium.current_window_handle
```
whimboo: Why do we return the current window handle and not the one for the newly opened window? This causes extra work for any consumer of this method.

foxpuppet/windows.py
whimboo: If the current window handle would have been cached we would safe a round-trip from Selenium to Geckodriver/Marionette and back, but that is not the case here. So leaving this check shouldn't give us performance wins.

foxpuppet/windows.py
jrbenny35: If the handle isn't there it raises a 'Browser undefined' message. Is that okay? I'd like for it to be a bit more specific.

foxpuppet/windows.py
jrbenny35: Currently switch_to could be called from another method and potentially act upon the current window, but I see what you mean. Is assuming the current window is always active correct however?

foxpuppet/browser/window.py
jrbenny35: From my tests when you open a new window, Firefox switches to it automatically.

whimboo: This is the focus but Selenium does not change the handle automatically. A test has to explicitly do so.

foxpuppet/windows.py
whimboo: We might have to customize this exception given that foxpuppet has to handle more than a browser window only.

foxpuppet/windows.py
jrbenny35: Looking at this again, I think you are right. It is unnecessary code. Selenium's switch_to works well.

whimboo: Keep in mind that active does not necessarily mean focused. Its the handle Marionette/Selenium interacts with.

jrbenny35:
Trying to focus a window from a list and not passing it into the function is a bit of a challenge for me. I don't see how to go from windows[1] and pass that into focus, which needs to switch to that handle.

whimboo:
Eg. windows[1] will return an instance of a Window class. The window class itself has a focus method, which then calls focus itself with the handle as attached via a property to the Window instance.

jrbenny35:
@whimboo so in that case we need a BaseWindow class like puppeteer, that way we can separate multiple window functions from a single window correct?

whimboo:
That's the `Window` class we talked about yesterday. But yes, this is a base class for any other kind of window type.

jrbenny35:
Just to be clear, this wasn't only me and dave. Most of these changes are the realization of what our last meeting was about.

foxpuppet/browser/window.py
```diff
@@ -2,43 +2,76 @@
...
-class BrowserWindow(object):
+class BrowserWindow(BaseWindow):
     _file_menu_button_locator = (By.ID, 'file-menu')
     _file_menu_private_window_locator = (By.ID, 'menu_newPrivateWindow')
     _file_menu_new_window_button_locator = (By.ID, 'menu_newNavigator')
     _nav_bar_locator = (By.ID, 'nav-bar')
     _tab_browser_locator = (By.ID, 'tabbrowser-tabs')
 
     def __init__(self, selenium, *args, **kwargs):
+        super(BrowserWindow, self).__init__(selenium)
```
whimboo: We should forward *args and **kwargs.

foxpuppet/browser/window.py
```diff
@@ -2,43 +2,76 @@
...
         self.selenium = selenium
```
whimboo: This should be done by the base window class?

foxpuppet/browser/window.py
```diff
@@ -2,43 +2,76 @@
...
+                """, self.window_element)
+
+    @property
+    def window_element(self):
+        """ Returns the inner DOM window element.
+
+        :returns: DOM window element.
+        """
+
+        return self.selenium.find_element(By.CSS_SELECTOR, ':root')
```
whimboo: This also has to be a property of the base window class.

whimboo: Also it returns the root element of a window but not a DOM window element.

foxpuppet/browser/window.py
```diff
@@ -2,43 +2,76 @@
...
+    """
+
+    def __init__(self, selenium, handles_before):
+        self.selenium = selenium
+        self.handles_before = handles_before
+
+    def __call__(self, selenium):
+        for handle in self.selenium.window_handles:
+            if handle not in self.handles_before:
+                return handle
```
whimboo: I assume this is something you both came up during the work week? In any case this is also base window related.

foxpuppet/windows.py
```diff
@@ -2,44 +2,50 @@
...
     def __init__(self, selenium):
         self.selenium = selenium
+        self.windows = None
+
+        self.populate_windows()
+
+    def populate_windows(self):
+        self.windows = []
+        for handle in self.selenium.window_handles:
+            self.windows.append(BaseWindow(self.selenium, handle))
```
whimboo: Why do we only instantiate base window classes here? Is this just temporarily?

Also why is this not a property? You always have to retrieve a new list when you want to access WindowManager.windows.

foxpuppet/windows.py
```diff
@@ -2,44 +2,50 @@
...
+
+
+class BaseWindow(object):
```
whimboo: We wanted to call this just "Window".

foxpuppet/windows.py
```diff
@@ -2,44 +2,50 @@
...
     @property
-    def all(self):
-        """Returns all current window handles"""
-        return self.selenium.window_handles
+    def handle(self):
+        return self._handle
+
+    @handle.setter
+    def handle(self, value):
+        self._handle = value
```
whimboo: If there is no specific code we have to execute to get/set a handle, just get rid of the getter/setter. It's useless.

foxpuppet/windows.py
```diff
@@ -2,44 +2,50 @@
...
 
     @property
-    def current(self):
-        """Returns current window handle"""
-        return self.selenium.current_window_handle
+    def wait(self):
+        return WebDriverWait(self.selenium, 10)
```
whimboo: Why do we hard-code values?

foxpuppet/windows.py
```diff
@@ -2,44 +2,50 @@
...
-    def window_element(self):
-        """Returns the inner DOM window element.
+        self.selenium.switch_to.window(self.handle)
 
-        :returns: DOM window element.
-        """
+        with self.selenium.context('chrome'):
+            self.selenium.execute_script('window.focus();')
 
-        return self.selenium.find_element(By.CSS_SELECTOR, ':root')
```
whimboo: I dont really understand why this line and others have been moved to the browser window class.

setup.py
```diff
@@ -8,13 +8,13 @@
 
 setup(
     name='FoxPuppet',
-    version='0.1.0',
+    version='0.2.0',
```
whimboo: I don't think we have to bump our version number at this stage, or is there any planned release?

tests/test_browser_model.py
```diff
@@ -11,17 +10,25 @@ class TestBrowserModel(object):
...
         """Tests opening a new private browsing window via menu"""
         foxpuppet = FoxPuppet(selenium)
-        foxpuppet.browser.open_window(private=True)
-        assert len(foxpuppet.windows.all) == 2
-        assert foxpuppet.browser.is_private is False
-        foxpuppet.windows.focus(foxpuppet.windows.all[1])
-        assert foxpuppet.browser.is_private is True
+        wm = foxpuppet.window_manager
+        wm.windows[0].handle = selenium.current_window_handle
+        print(wm.windows[0].handle)
```
whimboo: This code is totally confusing. Why are you setting the handle of a window outside of the class?

tests/test_browser_model.py
```diff
@@ -11,17 +10,25 @@ class TestBrowserModel(object):
...
+        assert len(wm.windows) == 1
+        handle = foxpuppet.browser.open_window(private=True)
+        wm.populate_windows()
```
whimboo: This never has to be called manually.

foxpuppet/browser/window.py
jrbenny35: What exactly are you saying should be done by the base class?

foxpuppet/windows.py
jrbenny35: The design of this was to set up a base window for each current window open. Kinda like the create_window_instance class in puppeteer

foxpuppet/browser/window.py
jrbenny35: I think dave and wanted this to be its own module, which is easy enough to move.

foxpuppet/windows.py
jrbenny35: Why not Base Window? I think window is kind of abstract no?

foxpuppet/windows.py
jrbenny35: This could probably be changed. I wanted to use a wrapper as it was a bit neater. Self.wait.until() was the api design

tests/test_browser_model.py
jrbenny35: I will move the populate method so it shouldn't be called manually.

tests/test_browser_model.py
jrbenny35: Kinda going with the idea of returning the handle when you open a window.

foxpuppet/browser/window.py
whimboo: There is no need that each and every subclass has to set the selenium property. It should be done by the base window class only.

foxpuppet/windows.py
whimboo: This method in Firefox Puppeteer doesn't create BaseWindow instances only. It checks which type of window it is and automatically creates the correct class instance.

foxpuppet/browser/window.py
whimboo: Ok, so please be aware that the performance for this method is poor. Check for the length of the array to see if a new handle is present. Iterating over and over through the list of handles consumes lot of valuable cpu time.

foxpuppet/windows.py
whimboo: This is the name we decided on during our last Vidyo call.

foxpuppet/windows.py
whimboo: Personally I'm not sure if that gives us a valuable helper. I will let Dave comment on it.

tests/test_browser_model.py
whimboo: We should return an instance of the appropriate window class, not a handle.

foxpuppet/windows.py
jrbenny35: Actually nevermind, I did what I suggested heh.

foxpuppet/browser/window.py
jrbenny35: That makes sense

foxpuppet/browser/window.py
jrbenny35: Yeah its linear time. The more handles there are the longer it will take. I have to come up with something to verify that a new window has actually opened. This is dirty but it works for now I guess.

foxpuppet/windows.py
jrbenny35: Can we add this later? I don't want to get to far off track like the last PR... ;)

foxpuppet/windows.py
jrbenny35: Can we switch back to basewindow? I think when we expand and have different types it will make more sense.

foxpuppet/browser/window.py
```diff
@@ -2,12 +2,14 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from foxpuppet.windows import Windows
-
 from selenium.webdriver.common.by import By
 
+from foxpuppet.windows import BaseWindow
+from foxpuppet.windows import new_window_is_opened
+from foxpuppet.windows import WindowManager
```
davehunt: Multiple imports from the same module should be written as `from module import thing1, thing2`

foxpuppet/windows.py
```diff
@@ -2,44 +2,75 @@
...
     def __init__(self, selenium):
         self.selenium = selenium
-        self._handle = None
 
     @property
-    def all(self):
-        """Returns all current window handles"""
-        return self.selenium.window_handles
+    def windows(self):
+        return [BaseWindow(self.selenium, handle=handle)
```
davehunt: `handle` is not an optional argument, I think we can remove the keyword.

foxpuppet/windows.py
```diff
@@ -2,44 +2,75 @@
...
-        yield
-        WebDriverWait(self.selenium, 10).until(
-            lambda driver: len(handles_before) != len(driver.window_handles))
+    def __init__(self, selenium, handle, *args, **kwargs):
+        self.selenium = selenium
+        self._handle = handle
+
+    @property
+    def handle(self):
+        return self._handle
```
davehunt: No need for this, just name the attribute `handle` instead of `_handle`.

foxpuppet/windows.py
```diff
@@ -2,44 +2,75 @@
...
 
         return self.selenium.find_element(By.CSS_SELECTOR, ':root')
+
+    def wait(self, timeout=10):
+        """
+        Selenium wait method
+
+        :param timeout: Sets a custom timeout length. Defaults to 10 seconds.
+        """
+        return WebDriverWait(self.selenium, timeout)
```
davehunt: The idea we talked about was to create an attribute off the base window:

```python
self.wait = WebDriverWait(self.selenium, 10)
```

This will provide a convenient default wait object.

foxpuppet/windows.py
```diff
@@ -2,44 +2,75 @@
...
+        with self.selenium.context('chrome'):
+            self.selenium.execute_script('window.focus();')
+
+    def close(self):
+        """Closes the window"""
+        self.selenium.switch_to.window(self.handle)
+        self.selenium.close()
+
+
+class new_window_is_opened(object):
```
davehunt: I think we should put this into an `expected` module.

foxpuppet/windows.py
```diff
@@ -2,44 +2,75 @@
...
+    """
+
+    def __init__(self, selenium, handles_before):
+        self.selenium = selenium
+        self.handles_before = handles_before
+
+    def __call__(self, *args, **kwargs):
+        for handle in self.selenium.window_handles:
+            if handle not in self.handles_before:
+                return handle
```
davehunt: We should wait for the number of handles to increase by one, and return a window object.

tests/test_browser_model.py
```diff
@@ -11,17 +10,22 @@ class TestBrowserModel(object):
     def test_new_private_window(self, selenium):
         """Tests opening a new private browsing window via menu"""
         foxpuppet = FoxPuppet(selenium)
+        wm = foxpuppet.window_manager
```
davehunt: I would either call `foxpuppet.window_manager` each time, or name this variable `window_manager`. Naming it `wm` makes it hard to read.

tests/test_browser_model.py
```diff
@@ -11,17 +10,22 @@ class TestBrowserModel(object):
...
+        print(wm.windows)
+        assert len(wm.windows) == 1
         foxpuppet.browser.open_window(private=True)
-        assert len(foxpuppet.windows.all) == 2
-        assert foxpuppet.browser.is_private is False
-        foxpuppet.windows.focus(foxpuppet.windows.all[1])
-        assert foxpuppet.browser.is_private is True
+        assert len(wm.windows) == 2
+        wm.windows[1].focus()
```
davehunt: `open_window` should return the window object, so we should be able to do `new_window.focus()` here.

tests/test_browser_model.py
```diff
@@ -11,17 +10,22 @@ class TestBrowserModel(object):
...
+        assert foxpuppet.browser.is_private
```
davehunt: I feel that `foxpuppet.browser` should be the initial browser window, and shouldn't follow whatever is the currently focused window. We should therefore have an assert like `assert new_window is not foxpuppet.browser` and the private check should be `assert new_window.is_private`.

foxpuppet/windows.py
davehunt: We discussed last week just assuming all windows were browsers for now. Once we add another window type we'll need a function that determines the appropriate window type.

foxpuppet/windows.py
davehunt: I suggested simply `Window`, but on reflection I'm happy with `BaseWindow`. This is also what I've done with the notifications patch. I'm also tempted to suggest that `BaseWindow` should be an abstract base class, and that it should not be instantiated directly.

foxpuppet/windows.py
davehunt: I've commented on the open pull request. The idea is that there is a `wait` with a sensible timeout value provided by the base window. If another timeout value is needed then another `WebDriverWait` can be instantiated.

tests/test_browser_model.py
davehunt: This was so that we could write a unit test for the method.

foxpuppet/browser/window.py
```diff
@@ -16,29 +18,42 @@ class BrowserWindow(object):
 
     def __init__(self, selenium, *args, **kwargs):
         self.selenium = selenium
-        self._windows = Windows(selenium)
```
whimboo: Why don't you call BaseWindow's __init__() method here?

foxpuppet/browser/window.py
```diff
@@ -16,29 +18,42 @@ class BrowserWindow(object):
...
-        self.selenium.find_element(*self._file_menu_button_locator).click()
-        with self._windows.wait_for_new_window():
+        """Opens a new browser window
+
+        :param private: Optional parameter to open a private browsing window.
+                        Defaults to false.
+        """
+
+        with self.selenium.context('chrome'):
+            handles_before = self.selenium.window_handles
```
whimboo: This is most likely not doing the right thiing because window_handles != chrome_window_handles. Does Geckodriver pass through that property already?

foxpuppet/browser/window.py
```diff
@@ -16,29 +18,42 @@ class BrowserWindow(object):
...
                 self.selenium.find_element(
                     *self._file_menu_private_window_locator).click()
             else:
                 self.selenium.find_element(
                     *self._file_menu_new_window_button_locator).click()
-        self.selenium.set_context('content')
+
+        # Wait until the window has opened and a new handle is returned
+        self.wait(timeout=10).until(
+            new_window_is_opened(self.selenium, handles_before))
```
whimboo: Always use a message for until(). Otherwise only a generic Timeout exception without any details is thrown. Maybe we should use the default wait time too?

foxpuppet/browser/window.py
```diff
@@ -16,29 +18,42 @@ class BrowserWindow(object):
...
+
+        # Fill list with new windows and handles
+        WindowManager.windows
```
whimboo: Why do we have to do that here? We need it when some code needs this list. Also why don't we return the new window instance here?

foxpuppet/windows.py
whimboo: So for now we only want to return BaseWindow instances? We can implement returning the correct instance in a follow-up patch.

foxpuppet/windows.py
whimboo: That's already something I mentioned in my last review. Maybe Benny missed it to change.

foxpuppet/windows.py
```diff
@@ -2,44 +2,75 @@
...
 
-    @contextmanager
-    def wait_for_new_window(self):
-        handles_before = self.selenium.window_handles
-        yield
-        WebDriverWait(self.selenium, 10).until(
-            lambda driver: len(handles_before) != len(driver.window_handles))
+    def __init__(self, selenium, handle, *args, **kwargs):
+        self.selenium = selenium
+        self._handle = handle
```
whimboo: To allow a propert MRO even this class should call the __init__ method from the super class.

foxpuppet/windows.py
```diff
@@ -2,44 +2,75 @@
...
 
     @property
     def window_element(self):
-        """Returns the inner DOM window element.
+        """ Returns the inner DOM window element.
 
         :returns: DOM window element.
         """
 
         return self.selenium.find_element(By.CSS_SELECTOR, ':root')
```
whimboo: Similar to other properties the strategy and value should be definded as class property.

foxpuppet/windows.py
whimboo: Now I see. Thanks.

foxpuppet/windows.py
whimboo: Similar here regarding my last review. I would suggest that we push an update once when all the comments have been addressed so we do not have an overlap?

foxpuppet/browser/window.py
jrbenny35: I knew it!

foxpuppet/windows.py
jrbenny35: Yeah @whimboo. Dave said we would add the different types of windows later.

foxpuppet/windows.py
jrbenny35: I like properties :(

foxpuppet/windows.py
whimboo: So shouldn't we return BrowserWindow instances instead as long as we only handle browser windows?

foxpuppet/browser/window.py
jrbenny35: 10 Seconds is the default. Is that fine?

foxpuppet/windows.py
jrbenny35: So John already had something like that with his wait_for_new_window context manager. Should I just use that code?

foxpuppet/browser/window.py
whimboo: If it's the default, there shouldn't be a need to have to specify it each time.

foxpuppet/browser/window.py
jrbenny35: I have to pass a handle to it, and I don't see how I could pass in a handle before opening the new window. It would be the old handle. 

I could pass in the old one and then replace it just to satisfy the init.

foxpuppet/windows.py
jrbenny35: What do you mean exactly?

foxpuppet/browser/window.py
whimboo: Which list are you talking about here?

tests/test_browser_model.py
jrbenny35: More Java habits

foxpuppet/windows.py
whimboo: See https://github.com/mozilla/FoxPuppet/pull/38/files#diff-d18c16bb3da239dae97766c71eed09e7R13

Something like _window_element = (By.CSS_SELECTOR, ':root')

Whereby we should rename all that given that is is the document_element.

jrbenny35:
How do you guys feel about setting selenium in every init? Would it be better to have the foxpuppet class set selenium?

foxpuppet/windows.py
jrbenny35: Maybe when we expand to allow different window types we can change it

foxpuppet/browser/window.py
jrbenny35: Oh! Good point

foxpuppet/browser/window.py
jrbenny35: I was talking about the list of window instance objects. I think I figured it out though

foxpuppet/windows.py
jrbenny35: Ah that's what I thought you meant but wanted to be sure.

whimboo:
> How do you guys feel about setting selenium in every init? Would it be better to have the foxpuppet class set selenium?

If we don't want to have this line for each and every sub class we could have a base meta class which is used by any of the FoxPuppet classes. I would let @davehunt decide about it.

foxpuppet/windows.py
whimboo: We only support Browser windows at the moment. Why should we return BaseWindow instances?

foxpuppet/browser/window.py
whimboo: There is no explicit update necessary. The only thing we have to ensure is that when foxpuppet.windows is called that we return the correct list of window instances.

davehunt:
> How do you guys feel about setting selenium in every init? Would it be better to have the foxpuppet class set selenium?

I don't fully understand. We should set the selenium in `BaseWindow.__init__` so we don't need to do it anywhere else. As we talked about in Hawaii, I would think of the model as being windows and regions, similar to our Selenium projects being pages and regions. Every region would have a relationship to a window, so regions would be able to access the selenium object via `self.window.selenium`, though we'd probably just set `self.selenium = self.window.selenium` in a `BaseRegion` object. This is outside the scope for this pull request though.

foxpuppet/windows.py
davehunt: It will still act like a property.

foxpuppet/windows.py
davehunt: You can, and we can review it. I don't think John's code returned a window object though. Also, I think it just checked that the length was greater. We expect it to increase by exactly one.

whimboo:
> Every region would have a relationship to a window, so regions would be able to access the selenium object via self.window.selenium, though we'd probably just set self.selenium = self.window.selenium in a BaseRegion object. This is outside the scope for this pull request though.

Oh, right. That would also work. So it's fine then. Thanks Dave!

foxpuppet/browser/window.py
davehunt: This is an interesting (and key) point. Perhaps it's not an issue until we need to support other window types, but we definitely need to know the answer to this question.

foxpuppet/windows.py
davehunt: I might favour `_root_element`, but we can revisit this later if needed.

davehunt:
Let us know when this is ready for another review @jrbenny35. I'm on PTO from Thursday 22nd, so if it's after then I won't get to it until the new year (I'm back on the 3rd).

foxpuppet/browser/window.py
whimboo: As long as there is a single tab open in a chrome window there will be no difference. But if you try to close a chrome window with more than 2 open tabs the above issue comes into play. Marionette also has a different close method for chrome windows which is called `self.close_chrome_windows`.

foxpuppet/windows.py
jrbenny35: It was a small change so I did it.

foxpuppet/windows.py
jrbenny35: Wouldn't it be bad to just return the last element in the selenium window list instead of going through and checking each one? I really don't see how to get a new element without iterating through all the handles.

davehunt: Use something like http://stackoverflow.com/a/3462160/154975 to get the difference between the lists. So long as the length of the new list is 1, return a window object based on the handle.

foxpuppet/windows.py
jrbenny35: Getting a bit of a circular import issue when trying to return BrowserWindow Instances from the WindowManager Object

davehunt: @jrbenny35 can you push your code so we can help?

jrbenny35:
I feel the window manager class should not be in the same py file as the BaseWindow Class.

foxpuppet/browser/window.py
```diff
@@ -2,43 +2,57 @@
...
 
-class BrowserWindow(object):
+class BrowserWindow(BaseWindow):
     _file_menu_button_locator = (By.ID, 'file-menu')
     _file_menu_private_window_locator = (By.ID, 'menu_newPrivateWindow')
     _file_menu_new_window_button_locator = (By.ID, 'menu_newNavigator')
     _nav_bar_locator = (By.ID, 'nav-bar')
     _tab_browser_locator = (By.ID, 'tabbrowser-tabs')
 
     def __init__(self, selenium, *args, **kwargs):
```
davehunt: You shouldn't need to call out the `selenium` argument, it will just be passed through. In fact, if this `__init__` doesn't do anything unique, you can remove it entirely.

foxpuppet/expected.py
```diff
@@ -0,0 +1,19 @@
...
+
+    :returns handle: The new window handle
+    """
+
+    def __init__(self, selenium, handles_before):
+        self.selenium = selenium
+        self.handles_before = handles_before
+
+    def __call__(self, *args, **kwargs):
+        return set(self.selenium.window_handles) - set(self.handles_before)
```
davehunt: This is returning a list. We should guard this with an if statement for when there is only one item in the list, and return a window object instead.

foxpuppet/windows.py
```diff
@@ -2,44 +2,86 @@
...
+        :returns: A list of Window Instances
+        """
+        return [BrowserWindow(self.selenium, handle=handle)
+                for handle in self.selenium.window_handles]
 
-    @property
-    def current(self):
-        """Returns current window handle"""
-        return self.selenium.current_window_handle
+    def create_window_instance(self, handle):
```
davehunt: Why do we need this? Can't we just create the window instance where it's needed? I can understand that we might need a window factory once we have more than one window type - is that what you're preparing for? If so, I would suggest a `WindowFactory` class with a `create` method. I'm fine with leaving this until we have another window type though.

foxpuppet/windows.py
```diff
@@ -2,44 +2,86 @@
...
-    def wait_for_new_window(self):
-        handles_before = self.selenium.window_handles
-        yield
-        WebDriverWait(self.selenium, 10).until(
-            lambda driver: len(handles_before) != len(driver.window_handles))
+
+class BaseWindow(object):
+    _document_element = (By.CSS_SELECTOR, ':root')
+
+    def __init__(self, selenium, handle=None, *args, **kwargs):
```
davehunt: Why is `handle` optional?

foxpuppet/windows.py
```diff
@@ -2,44 +2,86 @@
...
+        self.selenium = selenium
```
davehunt: This shouldn't be needed, just use `super`.

foxpuppet/windows.py
davehunt: This should be an attribute, not a method. See code example in earlier comment.

foxpuppet/windows.py
davehunt: This needs to be removed.

foxpuppet/windows.py
jrbenny35: But this is the base class. What would super do here?

davehunt: my mistake 😂

jrbenny35:
@whimboo  Looking at puppeteer, the create_window_instance method creates a BaseWindow class. Wouldn't it make more sense to create a base class for each window instead of creating specific classes for each window?

jrbenny35:
The problem I am having is that if I make the method create_window_instance return a BrowserWindow object, everything gets messed up because its now a circular import. BrowserWindow has things like open_window. Which creates a new window object.

davehunt:
> The problem I am having is that if I make the method create_window_instance return a BrowserWindow object, everything gets messed up because its now a circular import.

I don't see why we need `create_window_instance` at this point.

jrbenny35:
Well we need windows which does petty much the same thing as create_window_instance.

jrbenny35:
I hope you guys understand what I am saying. I think the right move is to return a ```BaseWIndow``` object and use the ```BrowserWindow``` class as a helper.

jrbenny35:
@davehunt what do you think about implementing a call back function to pass how the new window should be opened into the ```BaseWindow``` class? Looking at puppeteer, it's a pretty clean way to do it.

jrbenny35:
So I had to make the handle optional because calling the super method from the ```BrowserWindow``` class would always fail and say that I needed ```handle```

jrbenny35:
So looks like I wasn't the only one running into the circular dependency problem.

https://dxr.mozilla.org/mozilla-central/source/testing/marionette/puppeteer/firefox/firefox_puppeteer/ui/windows.py#19

jrbenny35:
We also need something or someway to get the window type. I don't see a way to do that in selenium, at least within the docs.

whimboo:
> So I had to make the handle optional because calling the super method from the BrowserWindow class would always fail and say that I needed handle

Sorry, but I do not understand this sentence. The super class of BrowserWindow is BaseWindow and it needs the handle. If BrowserWindow.__init__ doesn't handle it, remove the param from the list and it will be passed through via *args.

whimboo:
> We also need something or someway to get the window type. I don't see a way to do that in selenium, at least within the docs.

Marionette is simply doing the following:

https://dxr.mozilla.org/mozilla-central/rev/7083c0d30e75fc102c715887af9faec933e936f8/testing/marionette/driver.js#1083

So pretty easy to duplicate without having it to add to Geckodriver immediately.

davehunt:
> @davehunt what do you think about implementing a call back function to pass how the new window should be opened into the BaseWindow class? Looking at puppeteer, it's a pretty clean way to do it.

I am not a fan of callbacks, but I don't object to them when they're necessary. I'm not understanding why they're necessary here though. Aren't we continuing to increase the scope of this patch by doing this?

> We also need something or someway to get the window type.

Again, do we need this now? We only have one window type, so let's just use that. We can be smarter once there's another window type introduced.

whimboo:
> I am not a fan of callbacks, but I don't object to them when they're necessary. I'm not understanding why they're necessary here though. Aren't we continuing to increase the scope of this patch by doing this?

You are right here. We should really stop adding more and more. Please concentrate on only fixing the remaining issue. Anything else should come as a separate PR.

jrbenny35:
I am not adding more and more. What ever I am adding is trying to solve the problem of a circular dependency so It can create ```BrowserWindow``` objects instead of ```BaseWindow``` objects.

jrbenny35:
The call back removed the need for ```WindowManager``` within the ```BrowserWindow``` class, further abstracting it. I think it makes sense. The problem still occurs though.

foxpuppet/expected.py
```diff
@@ -0,0 +1,22 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+
+class new_browser_is_opened(object):
+    """
+    An expectation for checking that a new window is found after requesting
+    it to be opened
+
+    :returns handle: The new window handle
```
davehunt: This returns the `BrowserWindow` object.

foxpuppet/expected.py
```diff
@@ -0,0 +1,22 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+
+class new_browser_is_opened(object):
```
davehunt: Rename to `new_browser_window_is_opened`.

foxpuppet/foxpuppet.py
```diff
@@ -2,13 +2,15 @@
...
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from .browser.window import BrowserWindow
-from .windows import Windows
+
+from .windows import WindowManager
 
 
 class FoxPuppet(object):
+    """ Base class that sets up the api for interacting with the Firefox browser.
```
davehunt: 'Base class' suggests this class is extended.

foxpuppet/windows/__init__.py
```diff
@@ -0,0 +1,54 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support.wait import WebDriverWait
+
+
+class WindowManager(object):
```
davehunt: I know this is from my code, but I was being lazy by putting things into `__init__.py`. Could you create separate modules for `WindowManager` and `BaseWindow`.

foxpuppet/windows/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def __init__(self, selenium):
+        self.selenium = selenium
+
+    @property
+    def windows(self):
+        """
+        Sets all current window handles to appropriate window instances
+        :returns: A list of Window Instances
```
davehunt: A list of BrowserWindow objects.

foxpuppet/windows/browser.py
```diff
@@ -2,43 +2,59 @@
...
     _file_menu_private_window_locator = (By.ID, 'menu_newPrivateWindow')
     _file_menu_new_window_button_locator = (By.ID, 'menu_newNavigator')
     _nav_bar_locator = (By.ID, 'nav-bar')
     _tab_browser_locator = (By.ID, 'tabbrowser-tabs')
 
-    def __init__(self, selenium, *args, **kwargs):
-        self.selenium = selenium
-        self._windows = Windows(selenium)
+    def __init__(self, *args, **kwargs):
+        super(BrowserWindow, self).__init__(*args, **kwargs)
```
davehunt: We should be able to remove this entirely, as we're not doing anything different from the parent class.

foxpuppet/windows/__init__.py
jrbenny35: These should be in the foxpuppet folder, not windows correct?

foxpuppet/windows/browser.py
jrbenny35: ```self.document_element``` is from ```BaseWindow```. If we don't need to extend it we can import from it.

foxpuppet/windows/__init__.py
jrbenny35: Actually I got it, obviously it should be in the windows class.

foxpuppet/foxpuppet.py
```diff
@@ -3,11 +3,11 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
 
-from .windows import WindowManager
+from .windows.window_manager import WindowManager
```
davehunt: I think it would be nice for us to be able to import `WindowManager` directory from the `windows` module. We should be able to do this by adding the import to `windows/__init__.py`. I think we could do this with `BaseWindow` and maybe even `BrowserWindow` too. This is really just cosmetic, though. What do you think @whimboo?

jrbenny35: I was thinking that too.

whimboo: I'm always up for imports from folders. For those which are most often used we can even include them into the top-level __init__.py.

jrbenny35:
Updated everything to what was asked. I also removed the focus method as I didn't see us using it for a little.

foxpuppet/windows/browser.py
```diff
@@ -2,43 +2,59 @@
...
 
     @property
     def is_private(self):
         """Returns True if this is a Private Browsing window."""
 
-        self.selenium.set_context('chrome')
-        return self.selenium.execute_script("""
+        self.selenium.switch_to.window(self.handle)
```
whimboo: This line needs an abstraction in BaseWindow. Please add a method called `switch_to(self)' which switches to the window. On any other place we can simply use `self.switch_to()` then.

foxpuppet/windows/browser.py
```diff
@@ -2,43 +2,59 @@
...
+                """, self.document_element)
 
     def open_window(self, private=False):
-        self.selenium.set_context('chrome')
-        self.selenium.find_element(*self._file_menu_button_locator).click()
-        with self._windows.wait_for_new_window():
+        """Opens a new browser window
+
+        :param private: Optional parameter to open a private browsing window.
+                        Defaults to false.
```
whimboo: s/false/False

foxpuppet/windows/browser.py
```diff
@@ -2,43 +2,59 @@
...
                 self.selenium.find_element(
                     *self._file_menu_private_window_locator).click()
             else:
                 self.selenium.find_element(
                     *self._file_menu_new_window_button_locator).click()
-        self.selenium.set_context('content')
+
+        return self.wait.until(
+            new_browser_window_is_opened(self.selenium, handles_before),
+            message="No new browser window opened")
```
whimboo: As mentioned earlier, I'm not a bit fan of this logic. The handling of window handles should be transparent to the caller. So context manager or a callback sounds better alternatives.

tests/test_browser_model.py
```diff
@@ -3,25 +3,37 @@
...
+    assert not foxpuppet.browser.is_private
+
+
+def test_new_private_window(foxpuppet):
+    """Tests opening a new private browsing window via menu"""
+    assert len(foxpuppet.window_manager.windows) == 1
+    new_browser = foxpuppet.browser.open_window(private=True)
+    assert new_browser is not foxpuppet.browser
+    assert new_browser.is_private
+    assert len(foxpuppet.window_manager.windows) == 2
```
whimboo: Hm, where do you close this opened private browsing window?

foxpuppet/expected.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def __init__(self, selenium, handles):
+        self.selenium = selenium
+        self.handles = handles
+
+    def __call__(self, *args, **kwargs):
+        handles = list(set(self.selenium.window_handles) - set(self.handles))
+        if len(handles) == 1:
+            from .windows.browser import BrowserWindow
+            return BrowserWindow(self.selenium, handles[0])
```
davehunt: Agreed. I'd see us deprecating this in the future, which is why I was keen to have the window type mentioned in the name. We'd ideally have a generic expectation that returned the specific window type.

foxpuppet/windows/browser.py
davehunt: Let's keep this for now, and look into alternatives in a separate patch.

tests/test_browser_model.py
davehunt: pytest-selenium calls `selenium.quit` between each test, which closes all open windows, so we shouldn't need to explicitly close the window.

.travis.yml
```diff
@@ -35,6 +32,10 @@ matrix:
     - python: 3.5
       env: TOXENV=flake8
     - python: 3.5
+      env: TOXENV=coverage
+      addons:
+        firefox: latest
+    - python: 3.5
```
davehunt: Recent changes from master shouldn't be included. Did you rebase or merge?

foxpuppet/windows/__init__.py
```diff
@@ -0,0 +1,9 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from .browser import BrowserWindow
+from .window_manager import WindowManager
+
+
+__all__ = ['BrowserWindow', 'WindowManager']
```
davehunt: This line isn't required.

foxpuppet/windows/__init__.py
```diff
@@ -0,0 +1,9 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from .browser import BrowserWindow
```
davehunt: Add `BaseWindow` here too.

foxpuppet/windows/browser.py
```diff
@@ -2,43 +2,59 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from foxpuppet.windows import Windows
-
 from selenium.webdriver.common.by import By
 
+from foxpuppet.expected import new_browser_window_is_opened
+from foxpuppet.windows.base import BaseWindow
```
davehunt: We should import `BaseWindow` from the `windows` module. I also notice that we switch between relative and absolute imports. We can address this in a separate patch though.

foxpuppet/windows/browser.py
```diff
@@ -2,43 +2,59 @@
...
+        super(BrowserWindow, self).__init__(*args, **kwargs)
 
     @property
     def is_private(self):
         """Returns True if this is a Private Browsing window."""
 
-        self.selenium.set_context('chrome')
-        return self.selenium.execute_script("""
+        self.selenium.switch_to.window(self.handle)
+        with self.selenium.context('chrome'):
```
davehunt: Selenium provides class variables for contexts, we should switch to these in a separate patch: `self.selenium.CONTEXT_CHROME`

foxpuppet/windows/browser.py
jrbenny35: I can switch back. I didn't even notice heh

foxpuppet/windows/__init__.py
jrbenny35: So leaving them as imported but unused is fine?

.travis.yml
jrbenny35: I did a rebase to upstream/master

foxpuppet/windows/browser.py
jrbenny35: I am with henrik but another patch is appropriate.

tests/test_browser_model.py
whimboo: Oh, I always forget that we do this via Selenium.

.travis.yml
davehunt: Can you rebase again? This pull should only contain changes on top of current master.

jrbenny35: I think I had a conflict with my travis file and the one on master, so I had just took the one on master.

foxpuppet/windows/__init__.py
```diff
@@ -0,0 +1,6 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from .browser import BrowserWindow, BaseWindow # noqa
```
davehunt: BaseWindow is from `.base`, not `.browser`. I'll fix this before merging.

foxpuppet/windows/browser.py
```diff
@@ -2,43 +2,56 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from foxpuppet.windows import Windows
-
 from selenium.webdriver.common.by import By
 
+from ..expected import new_browser_window_is_opened
+from ..windows.base import BaseWindow
```
davehunt: I think we should use absolute imports. I mentioned we could do this in a later commit, but I'm not keen on this change going in.

davehunt: Also, this should be `from windows import BaseWindow`

foxpuppet/windows/browser.py
davehunt: Please use your new `switch_to` method.

foxpuppet/windows/browser.py
```diff
@@ -2,43 +2,56 @@
...
-        self.selenium.find_element(*self._file_menu_button_locator).click()
-        with self._windows.wait_for_new_window():
+        """Opens a new browser window
+
+        :param private: Optional parameter to open a private browsing window.
+                        Defaults to False.
+        """
+
+        handles_before = self.selenium.window_handles
+        self.selenium.switch_to.window(self.handle)
```
davehunt: Again, use your new method,

tests/test_browser_model.py
```diff
@@ -3,25 +3,37 @@
...
+
+
+def test_close_window(foxpuppet):
+    """Tests closing a window"""
+    assert len(foxpuppet.window_manager.windows) == 1
+    new_browser = foxpuppet.browser.open_window(private=False)
+    assert new_browser is not foxpuppet.browser
+    assert not new_browser.is_private
+    assert len(foxpuppet.window_manager.windows) == 2
```
davehunt: Please don't repeat asserts that are covered by other tests. This test only needs to check that we can close a window.

tests/test_browser_model.py
```diff
@@ -3,25 +3,37 @@
...
+    new_browser.close()
+    assert len(foxpuppet.window_manager.windows) == 1
```
davehunt: You've added a `switch_to` method, which should also have a test.

foxpuppet/windows/browser.py
jrbenny35: So focpuppet.expected?

tests/test_browser_model.py
jrbenny35: What do you mean repeat asserts?

jrbenny35: You mean don't do the is_private assert?

foxpuppet/windows/browser.py
davehunt: Yes

tests/test_browser_model.py
davehunt: This comment was meant to be against `test_close_window`. There are asserts in there that duplicate the checks from this test. We only need to assert when coverage is new.

foxpuppet/windows/browser.py
davehunt: This should be `from foxpuppet.windows import BaseWindow`

tests/test_browser_model.py
```diff
@@ -3,25 +3,42 @@
...
+def test_initial_browser_window(foxpuppet):
+    """Tests initial state of browser windows"""
+    assert len(foxpuppet.window_manager.windows) == 1
+    assert foxpuppet.browser is not None
+    assert not foxpuppet.browser.is_private
+
+
+def test_new_private_window(foxpuppet):
+    """Tests opening a new private browsing window via menu"""
+    assert len(foxpuppet.window_manager.windows) == 1
```
davehunt: This assert is covered by `test_initial_browser_window` and not needed here.

tests/test_browser_model.py
```diff
@@ -3,25 +3,42 @@
...
+    new_browser = foxpuppet.browser.open_window(private=True)
+    assert new_browser is not foxpuppet.browser
+    assert new_browser.is_private
+    assert len(foxpuppet.window_manager.windows) == 2
+
+
+def test_open_new_window(foxpuppet):
+    """Tests opening a new window via menu"""
+    assert len(foxpuppet.window_manager.windows) == 1
```
davehunt: This assert is covered by `test_initial_browser_window` and not needed here.

tests/test_browser_model.py
```diff
@@ -3,25 +3,42 @@
...
+    new_browser = foxpuppet.browser.open_window(private=False)
+    assert new_browser is not foxpuppet.browser
+    assert not new_browser.is_private
+    assert len(foxpuppet.window_manager.windows) == 2
+
+
+def test_close_window(foxpuppet):
+    """Tests closing a window"""
+    new_browser = foxpuppet.browser.open_window(private=False)
```
davehunt: You can omit `private=False` as this is the default behaviour, and it's not significant to this test.

tests/test_browser_model.py
```diff
@@ -3,25 +3,42 @@
...
+    new_browser.close()
+    assert len(foxpuppet.window_manager.windows) == 1
+
+
+def test_switch_to(foxpuppet):
+    """Test Switch to function"""
+    foxpuppet.browser.open_window(private=False)
```
davehunt: You can omit `private=False` as this is the default behaviour, and it's not significant to this test.

tests/test_browser_model.py
```diff
@@ -3,25 +3,42 @@
...
+
+    # Switch to originally window opened by pytest
+    foxpuppet.browser.switch_to()
+    assert foxpuppet.browser.handle == foxpuppet.selenium.current_window_handle
```
davehunt: I'd like to avoid encouraging use of `foxpuppet.selenium` (perhaps it should be `_selenium`). You can access the `selenium` object by including it in your test method arguments. I would also add a negative check on the window handle after opening the new window, to avoid a potential false positive.

jrbenny35: Heh, I was thinking that too.

jrbenny35:
Hey @davehunt, what do you mean by a problematic rebase? Is there something that I am missing?

davehunt:
> what do you mean by a problematic rebase? Is there something that I am missing?

This is what we were discussing on IRC yesterday. If you look at the commits for this pull you will see many commits repeated up to three times. You'll also see that there are commits from the master branch included in the pull request. I'm not even sure if an attempt to merge this pull would be successful, but it would almost certainly have undesired consequences. I'm okay with there being multiple commits, because I can squash these when I merge, but the commits do need to be in the correct order.

PR #39, opened 2016-12-01
Attempt to run tests on multiple Firefox versions

davehunt:
This should help us to determine and avoid regressing the versions of Firefox we support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,6 @@ python:
 
 sudo: "required"
 
-addons:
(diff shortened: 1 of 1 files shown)


davehunt:
Awesome, the tests pass in all Firefox versions. I wonder how long we can keep that up!

@jrbenny35 or @whimboo available to review?

jrbenny35:
I predict 100% uptime on all versions 😂

Issue #8, opened 2016-08-16
Notifications


silne30:
Implement Notifications Window

davehunt:
Initial notification support landed in #40.

Issue #13, opened 2016-08-16
Implement AddOnProgressNotification


silne30:

davehunt:
Initial notification support landed in #40. This specific notification wasn't include it, but let's close this issue and add support for this notification when there is a need.

Issue #11, opened 2016-08-16
Implement AddOnInstallCompleteNotification


silne30:

davehunt:
Initial notification support landed in #40.

Issue #10, opened 2016-08-16
Implement AddOnInstallConfirmationNotification


silne30:

davehunt:
Initial notification support landed in #40.

PR #40, opened 2016-12-01
Add initial support for notifications

davehunt:
In the end I retained a lot of the API decisions we made in puppeteer. I've also added a lightweight webserver that our tests can use if they need as a couple of these tests needed to be served an add-on. @whimboo or @jrbenny35 could you comment/review?


davehunt:
Fixes #8, #9, #10, #11, #12, and #13

jrbenny35:
Is there a reason it passed on nightly only?

davehunt:
> Is there a reason it passed on nightly only?

I'm sure there is, but I don't know why. This will take some investigation, but it's most likely related to how Travis is launching Firefox. I will look into it when I get time, but I'm still keen to hear feedback on the patch.

tests/test_notifications.py
```diff
@@ -0,0 +1,53 @@
...
+def test_wait_for_no_notification_timeout(firefox, notification):
+    """Wait for no notification when one is shown"""
+    with pytest.raises(TimeoutException) as excinfo:
+        firefox.wait_for_notification(None)
+    assert 'Unexpected notification shown' in str(excinfo.value)
+
+
+def test_notification_with_origin(firefox, webserver, notification):
+    """Trigger a notification with an origin"""
+    assert '{0.host}:{0.port}'.format(webserver) in firefox.notification.origin
```
whimboo: Nice! That formatting (with properties in the brackets)  is new for me. Learned something new.

tests/webserver.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def start(self):
+        self.thread.start()
+
+    def stop(self):
+        self.server.shutdown()
+        self.thread.join()
+
+    def url(self, path='/'):
+        return 'http://{0.host}:{0.port}/{1}'.format(self, path)
```
whimboo: I assume this module is a copy from another project and is tested well.

foxpuppet/browser/notifications.py
```diff
@@ -0,0 +1,103 @@
...
+        :returns: The notification origin.
+        """
+        with self.selenium.context('chrome'):
+            return self.root.get_attribute('origin')
+
+    def close(self):
+        """Close the notification."""
+        with self.selenium.context('chrome'):
+            self.root.find_anonymous_element_by_attribute(
+                'anonid', 'closebutton').click()
```
whimboo: Can you put this strategy on top similar to all the other element definitions? Maybe it would also be good to have a close button property on that class. Similar for other elements in that class.

foxpuppet/browser/window.py
```diff
@@ -3,22 +3,83 @@
...
+        try:
+            with self.selenium.context('chrome'):
+                root = self.selenium.find_element(*self._notification_locator)
+                _id = root.get_attribute('id')
+                return _map.get(_id, BaseNotification)(self, root)
+        except NoSuchElementException:
+            return None  # no notification is displayed
+
+    def wait_for_notification(self, notification_class=BaseNotification,
+                              timeout=5):
```
davehunt: I think @jrbenny35 is working on this, so I can add a TODO to fix this up when I can simply use `self.wait` with the default timeout.

setup.py
```diff
@@ -13,8 +13,10 @@
     long_description=open('README.rst').read(),
     url='https://github.com/mozilla/FoxPuppet',
     license='MPL2',
-    packages=['foxpuppet'],
+    packages=[
+        'foxpuppet',
+        'foxpuppet.browser'],
```
davehunt: You can discover modules, but as I've seen issues then trying to exclude test modules, etc.

tests/test_notifications.py
```diff
@@ -0,0 +1,53 @@
...
+def firefox_options(firefox_options):
+    firefox_options.add_argument('-foreground')
+    return firefox_options
+
+
+@pytest.fixture
+def notification(firefox, webserver, selenium):
+    selenium.get(webserver.url())
+    selenium.find_element_by_link_text('restartless_addon_signed.xpi').click()
+    return firefox.wait_for_notification(AddOnInstallBlocked)
```
davehunt: This isn't really a callback, but we're removing the context manager in the new window code in pull #38.

tests/webserver.py
davehunt: Why do you assume that? 😝 It's inspired by a number of different projects, but just a minimal webserver that will run in a separate thread.

foxpuppet/browser/notifications.py
davehunt: This is a little different because the strategy is not an argument, but hardcoded into the method `find_anonymous_element_by_attribute`. I could use something like  `_close_button_anonymous_id_attribute`, but this would increase maintenance rather than decrease it. I'm open to suggestions.

davehunt:
@whimboo ready for review again

setup.py
whimboo: You mean even with [find_packages](http://setuptools.readthedocs.io/en/latest/setuptools.html#using-find-packages)?

foxpuppet/browser/notifications.py
whimboo: Oh, you cannot use find_element() via GeckoDriver? In Marionette this works, so worth a try? See its [unit tests](https://dxr.mozilla.org/mozilla-central/source/testing/marionette/harness/marionette_harness/tests/unit/test_anonymous_content.py#68) which I recently updated. If that really doesn't work we should fix GeckoDriver IMHO.

davehunt: Not currently. Mind if I made this a TODO and reference a bug for now?

setup.py
davehunt: Yeah, I looked into it a while back and think I determined it was due to https://bitbucket.org/pypa/wheel/issues/99/cannot-exclude-directory

foxpuppet/browser/notifications.py
whimboo: If it doesn't work with Geckodriver lets file an issue there - maybe we also need webdriver? You can reference it for sure.

whimboo: @andreastt maybe you have some feedback for us for the above topic?

setup.py
whimboo: Ok, and the workaround to create the wheel from the sdist package is undesirable? If that is the case we could introduce find_packages later when the issue has been fully fixed.

tests/webserver.py
andreastt: You might want to consider using wptserve instead and save yourself a lot of trouble in the future. But don’t let this block you.

tests/webserver.py
```diff
@@ -0,0 +1,42 @@
...
+
+class WebServer(object):
+
+    def __init__(self, host='', port=8000):
+        self.server = HTTPServer((host, port), SimpleHTTPRequestHandler)
+        self.thread = threading.Thread(target=self._start_server)
+        self.thread.daemon = True
+
+    def _start_server(self):
+        os.chdir(os.path.join(os.path.dirname(__file__), 'web'))
```
andreastt: You probably want to override `translate_path` on `SimpleHTTPRequestHandler` instead of changing the CWD.

(Again, I think you’d be better off using wptserve for this.)

tests/webserver.py
andreastt: This will return "`http://127.0.0.1:1234//` if called with the default `path` argument. I think you have a superfluous `/` in the string format.

tests/test_notifications.py
```diff
@@ -0,0 +1,53 @@
...
+from selenium.common.exceptions import TimeoutException
+
+from foxpuppet.browser.notifications import (
+    BaseNotification,
+    AddOnInstallBlocked)
+
+
+@pytest.fixture
+def firefox_options(firefox_options):
+    firefox_options.add_argument('-foreground')
```
andreastt: Out of curiosity, what is this needed for?

foxpuppet/browser/notifications.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def install(self):
+        """Confirm add-on install."""
+        with self.selenium.context('chrome'):
+            self.root.find_element(*self._confirm_locator).click()
+
+
+class AddOnInstallComplete(BaseNotification):
+    """Add-on install complete notification."""
+    pass
```
andreastt: These passes are superfluous.

foxpuppet/browser/notifications.py
andreastt: @whimboo I believe this is coded the way it is due to restrictions in the Selenium Python binding’s API: last time I looked it was not easy to extend the permitted element location strategies. This may have changed, but would require an API change.

whimboo: So it means the Selenium bindings would need an addition to handle anonymous nodes for find_element?

andreastt: The anonymous element functions are implemented as extension commands with custom endpoints on geckodriver that are separate and disparate from the normal element retrieval endpoints.

This is in part due to typing differences as the data input taken for `FirefoxWebElement.anonymous_children` and `FirefoxWebElement.find_anonymous_element_by_attribute` is incompatible with the data input types expected for `RemoteWebElement.find_element`/`RemoteWebElement.find_elements`. The anonymous element methods allow input such as null and dictionaries (with some very funky behaviour in how this is treated, if you read the Marionette implementation), which doesn’t correspond to the (de)serialisation types defined by the Selenium remote server or the webdriver-rust library. In other words, trying to send `{"using": "anon attribute", "value": {}}` will cause a type check failure (probably an illegal attribute error) because the value type is not recognised.

The object serialisation for the extension commands in Selenium is found here: https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/firefox/webelement.py
And the mapping of the strings passed to URL routing is done here: https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/firefox/remote_connection.py

But also because the commands provide non-standard behaviour that other drivers don’t supply, there is a requirement that they are published separately, under a vendor-prefixed namespace:

* `/session/$sessionId/moz/xbl/$id/anonymous_children`
* `/session/$sessionId/moz/xbl/$id/anonymous_by_attribute`

It would probably be possible to smooth over this difference in the client binding by overriding `FirefoxWebElement.find_element`/`FirefoxWebElement.find_elements`, look at the strategy and then call the appropriate command; whether that be the standard element retrieval commands or the anonymous element extensions. But again, that would require an API change in the upstream library, and I’m not sure that’s in scope for this patch.

Anyway, I hope this illuminates the situation (-:

andreastt: The anonymous element functions are implemented as extension commands with custom endpoints on geckodriver that are separate and disparate from the normal element retrieval endpoints.

This is in part due to typing differences as the data input taken for `FirefoxWebElement.anonymous_children` and `FirefoxWebElement.find_anonymous_element_by_attribute` is incompatible with the data input types expected for `RemoteWebElement.find_element`/`RemoteWebElement.find_elements`. The anonymous element methods allow input such as null and dictionaries (with some very funky behaviour in how this is treated, if you read the Marionette implementation), which doesn’t correspond to the (de)serialisation types defined by the Selenium remote server or the webdriver-rust library. In other words, trying to send `{"using": "anon attribute", "value": {}}` will cause a type check failure (probably an illegal attribute error) because the value type is not recognised.

The object serialisation for the extension commands in Selenium is found here: https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/firefox/webelement.py
And the mapping of the strings passed to URL routing is done here: https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/firefox/remote_connection.py

But also because the commands provide non-standard behaviour that other drivers don’t supply, there is a requirement that they are published separately, under a vendor-prefixed namespace:

* `/session/$sessionId/moz/xbl/$id/anonymous_children`
* `/session/$sessionId/moz/xbl/$id/anonymous_by_attribute`

It would probably be possible to smooth over this difference in the client binding by overriding `FirefoxWebElement.find_element`/`FirefoxWebElement.find_elements`, look at the strategy and then call the appropriate command; whether that be the standard element retrieval commands or the anonymous element extensions. But again, that would require an API change in the upstream library, and I’m not sure that’s in scope for this patch.

Anyway, I hope this illuminates the situation (-:

whimboo: Ah I see. Thanks a lot for the detailed information about it. Looks like we have to keep it as it is right now.

tests/webserver.py
davehunt: I did try to use wptserve, but after the festive break I can no longer remember why I determined it wasn't suitable. I'll revisit and update here.

tests/test_notifications.py
davehunt: Without this, the popup notifications are not displayed.

tests/webserver.py
andreastt: You might find some inspiration for how to use it in https://github.com/mozilla/gecko-dev/blob/master/testing/marionette/harness/marionette_harness/runner/httpd.py and https://github.com/mozilla/gecko-dev/blob/master/testing/marionette/harness/marionette_harness/runner/serve.py.

tests/test_notifications.py
andreastt: OK, thanks. I think I’ve asked about this before, but what is a ‘popup notification’ in this case?

tests/webserver.py
davehunt: It might be the lack of support for Python 3... 😢

davehunt: Looks like we'd have to wait for https://github.com/w3c/wptserve/pull/93. I'll raise an issue to switch to wptserve in the future.

tests/test_notifications.py
whimboo: How does geckodriver or Selenium launch Firefox? -foreground should be used by default.

davehunt: I disagree. I feel that `-foreground` should not be needed, and the fact that it is could be considered a bug. @andreastt these are the 'doorhanger' type notifications that appear when you attempt to install an extension.

davehunt: By the way, Selenium with the legacy Firefox driver did pass `-foreground` by default, much to my frustration.

andreastt: We shouldn’t use `--foreground` by default because of a few different reasons:

- Automation of the web platform content does not require focus; this is a special case where it is fine to append `--foreground`
- The specification expects the driver to function without focus
- Many users run multiple Firefoxen on the same system and these could end up “competing” for focus

@davehunt Thanks!

stuartphilp: the underlying issue of having to ```-foreground```  to let notifications work seems like it should be fixed, i would think they should work regardless of whether ```-foreground``` is set. has that been discussed/filed?

@davehunt maybe a comment on this line to explain that as well might be useful

whimboo: Maybe we should file a bug and get feedback from Neil Deakin.

davehunt: I have raised [bug 1329939](https://bugzilla.mozilla.org/show_bug.cgi?id=1329939).

foxpuppet/windows/browser/notifications.py
```diff
@@ -0,0 +1,114 @@
...
+        :returns: The notification origin.
+        """
+        with self.selenium.context('chrome'):
+            return self.root.get_attribute('origin')
+
+    def close(self):
+        """Close the notification."""
+        with self.selenium.context('chrome'):
+            self.root.find_anonymous_element_by_attribute(
+                'anonid', 'closebutton').click()
```
whimboo: Don't we want to have the list of args as class definitions similar to other element declarations for find_element?

foxpuppet/windows/browser/notifications.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class AddOnInstallBlocked(BaseNotification):
+    """Add-on install blocked notification."""
+
+    def allow(self):
+        """Allow the add-on to be installed"""
+        with self.selenium.context('chrome'):
+            self.root.find_anonymous_element_by_attribute('anonid', 'button') \
+                .find_anonymous_element_by_attribute('anonid', 'button') \
```
whimboo: Having to search with the same arguments twice looks strange. Do we really have a button inside a button?

foxpuppet/windows/browser/notifications.py
```diff
@@ -0,0 +1,114 @@
...
+
+    @property
+    def addon_name(self):
+        """Provide access to the add-on name.
+
+        :returns: The add-on name.
+        """
+        with self.selenium.context('chrome'):
+            label = self.root.find_element(*self._addon_name_locator)
+            return label.get_attribute('value')
```
whimboo: It looks like that this needs get_property(). Can you please check? get_attribute() will only have a fallback to get_property() until Firefox 55.

foxpuppet/windows/browser/notifications.py
```diff
@@ -0,0 +1,114 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from abc import ABCMeta
+
+from selenium.webdriver.common.by import By
+
+
+class BaseNotification(object):
```
whimboo: Maybe we just call this `Notification'?

foxpuppet/windows/browser/notifications.py
```diff
@@ -0,0 +1,114 @@
...
+        'addon-install-confirmation-notification': AddOnInstallConfirmation,
+        'addon-install-complete-notification': AddOnInstallComplete,
+        'addon-install-restart-notification': AddOnInstallRestart,
+        'addon-install-failed-notification': AddOnInstallFailed,
+        'addon-progress-notification': AddOnProgress,
+    }
+
+    def create(self, window, root):
+        _id = root.get_attribute('id')
+        return self.NOTIFICATIONS.get(_id, BaseNotification)(window, root)
```
whimboo: I would propose that you move this create method into the base notification class and make it static. That way we can simply do `notification = Notification.create(window, element)`. Also the available notification types would be accessible directly without having to use a different class.

foxpuppet/windows/browser/window.py
```diff
@@ -0,0 +1,104 @@
...
+        """Waits for the specified notification to be displayed.
+
+        :param notification_class: Optional, the notification class to wait
+         for. If `None` is specified it will wait for any notification to be
+         closed. Defaults to `BaseNotification`.
+        :param timeout: Optional, how long to wait for the expected
+         notification. Defaults to 5 seconds.
+        """
+        # TODO use self.wait with the default timeout and remove this argument
+        wait = WebDriverWait(self.selenium, timeout=timeout)
```
whimboo: Hasn't this been landed yet?

foxpuppet/windows/browser/window.py
```diff
@@ -0,0 +1,104 @@
...
+                message = 'No notification was shown.'
+            else:
+                message = '{0} was not shown.'.format(
+                    notification_class.__name__)
+            wait.until(
+                lambda _: isinstance(self.notification, notification_class),
+                message=message)
+            return self.notification
+        else:
+            message = 'Unexpected notification shown.'
```
whimboo: You can remove this line.

.travis.yml
```diff
@@ -38,49 +53,65 @@ matrix:
...
+        apt:
+          packages:
+            - jwm
         firefox: latest
     - python: 3.5
       env: FIREFOX=esr
       addons:
+        apt:
+          packages:
+            - jwm
```
davehunt: I tried that first and it didn't work. I can ask/look around though.

foxpuppet/windows/__init__.py
```diff
@@ -3,5 +3,5 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
 from foxpuppet.windows.base import BaseWindow  # noqa: F401
-from foxpuppet.windows.browser import BrowserWindow  # noqa: F401
+from foxpuppet.windows.browser.window import BrowserWindow  # noqa: F401
```
davehunt: This is the `__init__.py` file, so we can use `from foxpuppet.windows import BrowserWindow`. Did you mean something else?

foxpuppet/windows/__init__.py
```diff
@@ -3,5 +3,5 @@
...
 from foxpuppet.windows.manager import WindowManager  # noqa: F401
```
davehunt: Which code is this reverting?

foxpuppet/windows/browser/notifications.py
davehunt: I followed the convention from `BaseWindow`. Is there a reason to break from convention here? This is also explicitly an abstract base class.

foxpuppet/windows/browser/notifications.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class BaseNotification(object):
+    """Abstract base class for any kind of notification."""
+
+    __metaclass__ = ABCMeta
+
+    def __init__(self, window, root):
+        self.window = window
+        self.root = root
```
davehunt: This indicates the root element for the notification. I'd like to introduce the `Region` concept, which `BaseNotification` would extend from. Using `parent` would be confusing as that's relative and we want this to be absolute.

foxpuppet/windows/browser/notifications.py
davehunt: Yes, but the reason I haven't here is because the strategy cannot by passed as an argument via Selenium, and we're restricted to using `find_anonymous_element_by_attribute`.

foxpuppet/windows/browser/notifications.py
davehunt: We really do. 😄

foxpuppet/windows/browser/window.py
davehunt: oops, thanks!

foxpuppet/windows/__init__.py
whimboo: Sorry, I was referring to foxpuppet/windows/browser/__init__.py.

foxpuppet/windows/__init__.py
whimboo: I thought to remember that you decided against the usage of `#noqa` and to skip this warning via the linter settings.

foxpuppet/windows/browser/notifications.py
whimboo: Whatever works best. As said earlier nothing is set in stone yet. This was just a thinking... I find it's looking superfluous.

foxpuppet/windows/browser/notifications.py
whimboo: Ok.

foxpuppet/windows/browser/notifications.py
whimboo: Well, we can still do `close_button = ('anonid', 'closebutton')` and pass it through this function. This would be a +1 for parity.

foxpuppet/windows/browser/notifications.py
whimboo: So what kinds of buttons are those? Especially the wrapper one.

whimboo:
The latest commit failed for Travis. Can you please have a look?

foxpuppet/windows/__init__.py
davehunt: I added the specific warning to ignore for each line, we wouldn't want to do this globally.

foxpuppet/windows/__init__.py
davehunt: I'm not sure I see the advantage in doing that?

davehunt:
Okay, so we're back to 100% coverage and a passing Travis CI build!

@whimboo: I would like your thoughts on https://github.com/mozilla/FoxPuppet/pull/40/commits/99b4424f3fcfa7f3cde86ed22a92f9b73954cb76, where I've introduced logic based on the version of Firefox due to notifications changing in Nightly. It was inevitable that this would happen, and if we were in mozilla-central we could just rely on the branches to avoid such logic, but then I feel it's worth trying to distribute a single package that supports Firefox release -> Nightly. I'm sure we could refine how we check the version though.

Also, I've disabled e10s in https://github.com/mozilla/FoxPuppet/pull/40/commits/4b3c601efa9f27d0a0133947dc445bce35761605, but I'd rather not commit this. I'd like to hear your thoughts - should we run all tests both with e10 enabled and with it disabled? This would be relatively easy to do, but would double the number of tests. If we leave it up to Firefox to decide, we might at least need to determine if it's enabled for FoxPuppet's tests.

foxpuppet/windows/browser/notifications/addons.py
```diff
@@ -39,12 +46,22 @@ def addon_name(self):
     def cancel(self):
         """Cancel add-on install."""
         with self.selenium.context(self.selenium.CONTEXT_CHROME):
-            self.root.find_element(*self._cancel_locator).click()
+            if self.window.firefox_version >= LooseVersion('53'):
+                # Firefox >= 53
```
jrbenny35: Could these comments be a bit more descriptive? Just thinking from the documentation perspective.

foxpuppet/windows/browser/notifications/addons.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            label = self.root.find_element(*self._addon_name_locator)
+            return label.get_property('value')
+
+    def cancel(self):
+        """Cancel add-on install."""
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            if self.window.firefox_version >= LooseVersion('53'):
+                # Firefox >= 53
```
jrbenny35: Could these comments be a bit more descriptive? Just thinking from a docs point of view.

davehunt: These comments would not be present in the documentation. I suspect we would make them more verbose though, if this is the approach we go with. I'm currently waiting on Henrik's thoughts, as this is really just enough to get something to work at the moment.

foxpuppet/windows/__init__.py
whimboo: Important classes under `foxpuppet.windows.browser.*` we should put into the __init__.py file for an easier import. For the window classes themselve I could even imagine to add them all to `foxpuppet/windows/__init__.py`, so we could do `from foxpuppet.windows import BrowserWindow`.

whimboo:
> @whimboo: I would like your thoughts on https://github.com/mozilla/FoxPuppet/pull/40/commits/99b4424f3fcfa7f3cde86ed22a92f9b73954cb76, where I've introduced logic based on the version of Firefox due to notifications changing in Nightly. It was inevitable that this would happen, and if we were in mozilla-central we could just rely on the branches to avoid such logic, but then I feel it's worth trying to distribute a single package that supports Firefox release -> Nightly. I'm sure we could refine how we check the version though.

Using the returned capability is fine. I haven't used `LooseVersion` yet, so I would trust you on its support for various kinds of builds, but as I see we only need the major version anyway. As mentioned already in another commit, I would still like to see the ui elements for anonymous nodes being defined similar to those which use `find_element()`. That way we could differentiate between DOM changes for element retrieval and actual application logic changes.

When we will have multiple branches later it should be easy for us to adapt the code for older trains.

Also, I've disabled e10s in https://github.com/mozilla/FoxPuppet/pull/40/commits/4b3c601efa9f27d0a0133947dc445bce35761605, but I'd rather not commit this. I'd like to hear your thoughts - should we run all tests both with e10 enabled and with it disabled? This would be relatively easy to do, but would double the number of tests. If we leave it up to Firefox to decide, we might at least need to determine if it's enabled for FoxPuppet's tests.

Could it be that the restartless add-on you are testing doesn't have the flag set for e10s support? Please have a look. I cannot believe that a restart is necessary otherwise.

.travis.yml
davehunt: I think the issue is that we want to override specific addons in the matrix. By specifying any addons in the matrix I believe we're replacing any defaults, so this is the only way to do what we need for now.

foxpuppet/windows/__init__.py
davehunt: @whimboo this is what we have already.. `from foxpuppet.windows import BrowserWindow` works exactly because of this line that you've commented on. We could move it to `foxpuppet/windows/browser/__init__.py` but we'd still also need `from foxpuppet.windows.browser import BrowserWindow` in this file. By making the change here, it's only one line.

whimboo: Yes, my failure. All fine.

foxpuppet/region.py
```diff
@@ -0,0 +1,21 @@
...
+class Region(object):
+    """A region object.
+
+    Used as a base class for region objects.
+
+    :param window: Window object this region appears in.
+    :param root: element that serves as the root for the region.
+    :type window: :py:class:`~.windows.BaseWindow`
+    :type root: :py:class:`~selenium.webdriver.remote.webelement.WebElement`
+    """
```
whimboo: I assume the above docstring is for the class properties. Can you add the same for the __init__ method please?

foxpuppet/windows/browser/notifications/addons.py
```diff
@@ -0,0 +1,88 @@
...
+class AddOnInstallBlocked(BaseNotification):
+    """Add-on install blocked notification."""
+
+    def allow(self):
+        """Allow the add-on to be installed"""
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            if self.window.firefox_version >= 53:
+                # Notifications were restyled in Firefox 53
+                self.root.find_anonymous_element_by_attribute(
+                    'anonid', 'button').click()
```
whimboo: Great that we no longer have this strange `anonid=button` element twice.

foxpuppet/windows/browser/notifications/base.py
```diff
@@ -0,0 +1,46 @@
...
+
+class BaseNotification(Region):
+    """Abstract base class for any kind of notification."""
+
+    __metaclass__ = ABCMeta
+
+    @staticmethod
+    def create(window, root):
+        notifications = {}
+        _id = root.get_attribute('id')
```
whimboo: I would propose to use `get_property()` here to fetch the id. Not sure if it is exposed as DOM attribute at any time.

foxpuppet/windows/browser/window.py
```diff
@@ -53,5 +89,6 @@ def open_window(self, private=False):
                     *self._file_menu_new_window_button_locator).click()
 
         return self.wait.until(
-            new_browser_window_is_opened(self.selenium, handles_before),
+            expected.new_browser_window_is_opened(
+                self.selenium, handles_before),
```
whimboo: If not done yet please file an issue so we can discuss to make that easier to use. Having to push the former handles as parameter doesn't make that much sense to me. It should be all transparent in the method itself.

tox.ini
```diff
@@ -1,16 +1,11 @@
...
-passenv = DISPLAY
-deps = -rrequirements/tests.txt
-commands = pytest --driver Firefox {posargs}
-
-[testenv:coverage]
 passenv = DISPLAY TRAVIS TRAVIS_JOB_ID TRAVIS_BRANCH
 deps = -rrequirements/coverage.txt
 commands =
-    pytest --driver Firefox --cov {posargs}
+    pytest --driver=Firefox --cov --html=results/{envname}.html {posargs}
```
whimboo: Does that mean we have coverage results for each supported Firefox release now and get x-amount of emails per push?

.travis.yml
```diff
@@ -14,73 +14,97 @@ matrix:
...
         firefox: latest
-    - python: 2.7
-      env: FIREFOX=esr
-      addons:
-        firefox: latest-esr
+    # - python: 2.7
```
davehunt: I suspect the comments might be part of the Travis CI failure. I'm going to remove these sections entirely, but I can add an issue to add ESR support once 52.1.0 ships.

foxpuppet/region.py
davehunt: By default sphinx autogen will document classes, and not `__init__` methods. See http://www.sphinx-doc.org/en/1.5.2/ext/autodoc.html#confval-autoclass_content

foxpuppet/windows/browser/window.py
davehunt: I don't have an issue with this, nor can I see a way around it. Could you perhaps raise this issue with your proposal?

tox.ini
davehunt: The coveralls results should be merged so that we reach 100% coverage due to the different code-paths each version may take. You shouldn't get any emails from coveralls (I don't) and there should be just one comment on the pull request per push, which might trigger an email notification depending on your GitHub settings.

foxpuppet/region.py
whimboo: Ah, learned something new then. So in general you only document the class and can skip the init method. I assume the parameters you are using here should match the init?

foxpuppet/windows/browser/window.py
whimboo: Which is issue #65 now.

tox.ini
whimboo: I get a couple of emails from github about the coveralls results for each commit anyone is doing on this repo.

davehunt: Yeah, I saw there were two.. I wonder if it's because of how coveralls uses environment variables from Travis. Unfortunately there's no documentation on this as far as I could find. It might be worth experimenting with.

PR #41, opened 2016-12-09
Add details to package, including switching to setuptools_scm for versions

davehunt:


setup.py
```diff
@@ -2,19 +2,30 @@
...
-)
+setup(name='FoxPuppet',
+      use_scm_version=True,
+      description='Firefox user interface testing model for use with Selenium',
+      long_description=open('README.rst').read(),
+      author='Firefox Test Engineering',
+      author_email='firefox-test-engineering@mozilla.com',
+      url='https://github.com/mozilla/FoxPuppet',
+      packages=['foxpuppet'],
+      install_requires=['selenium>=3.0.1'],
```
jrbenny35: Can we use 3.0.2 as the least compatible? This includes the context manager John wrote.

whimboo:
Beside the comment from Benny all looks fine to me.

setup.py
davehunt: I haven't introduced that requirement in this patch. I feel the dependency should only be bumped alongside the patch that uses the new context manager.

whimboo: I agree with Dave here. Lets get this PR landed. All fine from my side Dave.

PR #50, opened 2017-01-09
Switch to Travis CI container based infrastructure

davehunt:
See https://docs.travis-ci.com/user/migrating-from-legacy/#Why-migrate-to-container-based-infrastructure%3F for reasons to migrate.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,6 @@ language: python
 python:
 - 2.7
 - 3.5
-
-sudo: "required"
-
 env:
   global:
     - DISPLAY=:99.0
@@ -63,9 +60,9 @@ matrix:
     - python: 3.5
 
 before_install:
-  - wget -O geckodriver.tar.gz https://github.com/mozilla/geckodriver/releases/download/v0.12.0/geckodriver-v0.12.0-linux64.tar.gz
-  - gunzip -c geckodriver.tar.gz | tar xopf -
-  - chmod +x geckodriver && sudo mv geckodriver /usr/local/bin
+  - wget -O /tmp/geckodriver.tar.gz https://github.com/mozilla/geckodriver/releases/download/v0.12.0/geckodriver-v0.12.0-linux64.tar.gz
+  - mkdir $HOME/geckodriver && gunzip -c /tmp/geckodriver.tar.gz | tar xopf - -C $HOME/geckodriver
+  - export PATH=$HOME/geckodriver:$PATH
   - firefox --version
   - geckodriver --version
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,9 +2,6 @@ language: python
 python:
 - 2.7
 - 3.5
-
-sudo: "required"
```
whimboo: I thought "sudo" has to be turned of to set its value to false. I cannot find anything about "required". Mind explaining why we need sudo?

.travis.yml
```diff
@@ -63,9 +60,9 @@ matrix:
     - python: 3.5
 
 before_install:
-  - wget -O geckodriver.tar.gz https://github.com/mozilla/geckodriver/releases/download/v0.12.0/geckodriver-v0.12.0-linux64.tar.gz
-  - gunzip -c geckodriver.tar.gz | tar xopf -
-  - chmod +x geckodriver && sudo mv geckodriver /usr/local/bin
+  - wget -O /tmp/geckodriver.tar.gz https://github.com/mozilla/geckodriver/releases/download/v0.12.0/geckodriver-v0.12.0-linux64.tar.gz
```
whimboo: Why cannot we download to the local dir? Or maybe even a sub directory which we could even add to the cache? We should also update to the 0.13.0 release of geckodriver.

.travis.yml
```diff
@@ -63,9 +60,9 @@ matrix:
...
+  - mkdir $HOME/geckodriver && gunzip -c /tmp/geckodriver.tar.gz | tar xopf - -C $HOME/geckodriver
```
whimboo: Why not calling "tar xfv geckodriver.tar.gz" directly?

.travis.yml
davehunt: I'm removing this line, the default is false and container based infrastructure.

.travis.yml
davehunt: Updating should be in a separate commit. I'd also argue that using cache should be taken care of separately.

.travis.yml
davehunt: This patch was more about switching to container based. I'm sure there are many other optimisations we can make independently. If you feel strongly I'll apply your suggestions in this pull, but as separate commits.

.travis.yml
davehunt: I think the default also depends on the age of the project, so projects that pre-date container based infrastructure actually default to true.

.travis.yml
whimboo: Sounds fine to me.

.travis.yml
whimboo: It's a small change and since we already touch this code... otherwise feel free to file a new issue and maybe a PR.

Issue #53, opened 2017-01-11
Upgrade support to Python 3.6


davehunt:
We should make sure that FoxPuppet supports the latest stable version of Python. Let's also change the Travis CI jobs to run against 3.6 and update the documentation.

jrbenny35:
Currently running into this issue. 

https://github.com/pypa/pypi-legacy/issues/322

If anyone could help that would be great.

davehunt:
@jrbenny35 did you resolve your issue? I see you opened a pull request...

PR #58, opened 2017-01-18
Updated tests to use selenium class variable. Updated to python 3.6

jrbenny35:
Solves #53

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 python:
 - 2.7
-- 3.5
+- 3.6
 env:
   global:
     - DISPLAY=:99.0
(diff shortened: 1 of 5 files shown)


jrbenny35:
Think I fast forwarded on accident. Maybe a squash and merge will keep this clean

jrbenny35:
You want me to close this branch then? I figured since they were both small that one PR was fine.

davehunt:
@jrbenny35 there's no need to close this pull request or remove the branch. You can just remove one of the fixes and push again. While it can be tempting to fix multiple small issues in a single pull request, it could mean fixes are delayed due to unrelated fixes needing further revisions. Also, we should never combine fixes for different issues into a single commit, as this makes the history harder to read, and makes it impossible to revert one fix without also reverting the others. These changes are very much not related, so should be separate commit and separate pull requests.

PR #61, opened 2017-01-20
Inital commit for the auto generation of the api docs

jrbenny35:
Just a rough draft really. 

@davehunt r?

Also after this I will do the User Guide.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1,0 +1,36 @@
+Developer Interface for FoxPuppet
+*********************************
+
+This part of the documentation describes the interfaces for using FoxPuppet.
+
+.. py:module:: foxpuppet.foxpuppet
+
+.. autoclass:: FoxPuppet
+   :inherited-members:
+
+
+Window Manager
+==============================
+
+.. py:module:: foxpuppet.windows
+
+.. autoclass:: WindowManager
+  :inherited-members:
+
+
+Base Window Model
+=================
+
+.. py:module:: foxpuppet.windows.base
+
+.. autoclass:: BaseWindow
+  :inherited-members:
+
+
+Browser Model
+=============
+
+.. py:module:: foxpuppet.windows.browser
+
+.. autoclass:: BrowserWindow
+  :inherited-members:
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -3,6 +3,9 @@ Developer Interface
 
 This part of the documentation describes the interfaces for using FoxPuppet.
 
+FoxPuppet
+---------
+
 .. py:module:: foxpuppet.foxpuppet
 
 .. autoclass:: FoxPuppet
@@ -11,6 +14,9 @@ This part of the documentation describes the interfaces for using FoxPuppet.
 
 Windows
 -------
+
+This directory contains all of the window types as well as the window manager.
+
 Window Manager
 ==============
 
@@ -19,8 +25,8 @@ Window Manager
 .. autoclass:: WindowManager
   :inherited-members:
 
-Browser
-=======
+Browser Window
+==============
 
 .. py:module:: foxpuppet.windows.browser
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1,5 +1,5 @@
-Developer Interface for FoxPuppet
-*********************************
+Developer Interface
+*******************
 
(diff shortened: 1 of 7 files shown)


docs/api.rst
```diff
@@ -0,0 +1,36 @@
...
+  :inherited-members:
+
+
+Browser Model
+=============
+
+.. py:module:: foxpuppet.windows.browser
+
+.. autoclass:: BrowserWindow
+  :inherited-members:
```
davehunt: The structure should be a heading for windows and then a section for each window implementation. We shouldn't need to document the abstract base class.

```
# Windows
## Window Manager
## Browser Window
```

docs/api.rst
```diff
@@ -0,0 +1,36 @@
...
+Base Window Model
+=================
+
+.. py:module:: foxpuppet.windows.base
+
+.. autoclass:: BaseWindow
+  :inherited-members:
+
+
+Browser Model
```
davehunt: Remove 'Model' from the headings.

foxpuppet/foxpuppet.py
```diff
@@ -8,7 +8,11 @@
 
 
 class FoxPuppet(object):
-    """ Class that sets up the api for interacting with the Firefox browser.
+    """
+        Class that sets up the interface for interacting with the Firefox
+        browser.
+
+        :param selenium: A selenium object
```
davehunt: We should use intersphinx and link the param type to the official Selenium documentation as we do with PyPOM: https://github.com/mozilla/PyPOM/blob/master/docs/conf.py#L41

foxpuppet/windows/browser.py
```diff
@@ -37,6 +41,8 @@ def open_window(self, private=False):
 
         :param private: Optional parameter to open a private browsing window.
                         Defaults to False.
+
+        :returns: A BrowserWindow object of the opened window.
```
davehunt: We should include the return types, linking to the relevant documentation.

foxpuppet/windows/manager.py
```diff
@@ -5,14 +5,23 @@
 
 class WindowManager(object):
 
+    """
+        A window manager that controls the creation of window objects
+        for interaction.
```
davehunt: I don't think this description is particularly useful. Perhaps say that it provides access to all open windows. The user doesn't need to know that it creates the window objects.

foxpuppet/windows/manager.py
```diff
@@ -5,14 +5,23 @@
...
         self.selenium = selenium
 
     @property
     def windows(self):
         """
-        Sets all current window handles to appropriate window instances
-        :returns: A list of BrowserWindow Instances
+        Sets all current windows to appropriate window instances.
+
+        :returns: A list of Window Instances
```
davehunt: Include the return types.

jrbenny35:
@davehunt Thanks for the review. When you say push to a branch on the mozilla fork, you mean create a branch on the Mozilla fork of foxpuppet?

docs/api.rst
jrbenny35: Would we have to say anything about the 'Windows' directory?

foxpuppet/windows/browser.py
jrbenny35: Linking to what specifically?

jrbenny35:
I pushed to a branch here but the docs aren't rendering the new changes. I can do a build locally and get them however.

davehunt:
I've enabled the sphix_auto_docs branch as a protected version here: http://foxpuppet.readthedocs.io/en/sphinx_auto_docs/api.html. As you can see the autogenerated docs are not showing up yet. Please close this pull, push only to the branch on the mozilla fork for now, and when you're ready open a new pull request from the new branch.

docs/api.rst
davehunt: I don't think so, users of a Python client will not think of this as a directory, but as a package/module.

foxpuppet/foxpuppet.py
jrbenny35: Is there a good example of this within a docstring? It would work within an rst file but not within a py file.

jrbenny35:
Closing.

PR #62, opened 2017-01-20
Rough draft of docs user guide

jrbenny35:
As the title states. I need a lot of feedback on this!


jrbenny35:
Hmmm, I know it seems like a lot but being new and having to read a lot of docs, I think its better to have very specific examples. I could remove some and condense it a bit but I feel we should be specific and not rely on the user to 'figure it out' as it were. Just my .02 cents/pence.

jrbenny35:
Should we also assume they are using pytest? I feel that we shouldn't.

jrbenny35:
@davehunt should I also push this branch to the mozilla repo directly?

jrbenny35:
Thinking about this all day and I agree. Remembering that the api will explain each function, maybe we could include examples there.

davehunt:
> Hmmm, I know it seems like a lot but being new and having to read a lot of docs, I think its better to have very specific examples. I could remove some and condense it a bit but I feel we should be specific and not rely on the user to 'figure it out' as it were. Just my .02 cents/pence.

At the moment we have very limited functionality, so a single example that opens a private browsing window and then closes it actually covers most of that. Determining if a window is private is unlikely to be something users will need (we needed it specifically for our own tests). I think it's fine to split the example up, but I think it's hard to follow a guide with bold headings for things that otherwise would flow together nicely. I'm sure there's a compromise here that we'll both be happy with. 😄

davehunt:
> Should we also assume they are using pytest? I feel that we shouldn't.

We shouldn't even assume that they're writing tests. 🤓

davehunt:
> should I also push this branch to the mozilla repo directly?

No, this is easy enough to view. I only wanted the other branch because it include generated docs that are otherwise hard to preview.

davehunt:
> Thinking about this all day and I agree. Remembering that the api will explain each function, maybe we could include examples there.

I'm fine with including usage examples in some of the docstrings. I believe we did this with PyPOM too.

docs/user_guide.rst
```diff
@@ -1,2 +1,44 @@
...
+Example
+-------
+
+This is an example for setting up FoxPuppet and running a simple check to see if the browser window is private::
+
+    from foxpuppet import FoxPuppet
+    from selenium.webdriver import Firefox
+
+
+    selenium = Firefox
```
davehunt: This should be `selenium = Firefox()`. Make sure you can copy/paste your examples into a Python interpreter.

docs/user_guide.rst
```diff
@@ -1,2 +1,44 @@
...
+    foxpuppet = FoxPuppet(selenium)
+
+    assert not foxpuppet.browser.is_private
+    foxpuppet.browser.open_window(private=True)
+    assert foxpuppet.browser.is_private
```
davehunt: Don't include asserts in your examples, as this presumes the user is writing a test. I would suggest opening a new window, and then closing it. We can add more to this example when we have more functionality. It would be neat to do something like type an address into the awesome bar, and then save it as a bookmark.

docs/user_guide.rst
```diff
@@ -1,2 +1,44 @@
...
+
+This test first checks the initial brower window to see if it is private. It then proceedes to open a private browser window
+and then it checks to see if that window is private.
+
+:Note: The initial browser window is automatically assigned to the :py:attr:`browser` attribute and is always available.
```
davehunt: 👍

docs/user_guide.rst
```diff
@@ -1,2 +1,44 @@
...
+
+Setup
+------
+
+FoxPuppet requires a Selenium :py:class:`~selenium.webdriver.remote.webdriver.WebDriver`
+object to be instantiated::
```
davehunt: This is covered in your example above - I don't think we need to repeat it.

docs/user_guide.rst
```diff
@@ -1,2 +1,44 @@
 User Guide
 ==========
+
+.. contents:: :depth: 3
+
+:Note: FoxPuppet uses Selenium as it's main driver. That means Selenium commands work
+       right along with FoxPuppet specific commands.
```
davehunt: This doesn't need to be a note.

jrbenny35:
It seems that the note about the initial window being available in browser isn't correct. It would be correct if we instantiated foxpuppet AFTER opening a window, say with ```selenium.get()```. I think most users will instantiated foxpuppet before opening a window or navigating to a page with selenium.

docs/user_guide.rst
```diff
@@ -1,2 +1,28 @@
 User Guide
 ==========
+
+.. contents:: :depth: 3
```
davehunt: We don't need a table of contents for such a short page.

docs/user_guide.rst
```diff
@@ -1,2 +1,28 @@
...
+
+This is an example for setting up FoxPuppet and running a simple check to see if the browser window is private::
+
+    from foxpuppet import FoxPuppet
+    from selenium.webdriver import Firefox
+
+
+    selenium = Firefox(
+        firefox_binary='path/to/firefox-bin'
+    )
```
davehunt: Let's simplify this to `selenium = Firefox()` to just pick up the default location.

docs/user_guide.rst
```diff
@@ -1,2 +1,28 @@
...
+    foxpuppet = FoxPuppet(selenium)
+
+    window = foxpuppet.browser.open_window(private=True)
+    window.close()
+    foxpuppet.browser.close()
```
davehunt: I don't think we should advise this, as users would probably not be able to recover from closing the original browser window. The `selenium.quit` call will take care of closing any remaining windows.

docs/user_guide.rst
```diff
@@ -1,2 +1,28 @@
 User Guide
 ==========
+
+.. contents:: :depth: 3
+
```
davehunt: I feel like we need at least one or two sentences of introduction before jumping in with an example. Perhaps repeat what FoxPuppet is for.

Issue #65, opened 2017-01-24
Better handle the wait for opening windows


whimboo:
Currently we have code like this:

```
       return self.wait.until(
            expected.new_browser_window_is_opened(
                self.selenium, handles_before),
```

Problem is that each consumer of the expected method would have to retrieve the handles before.

I propose that we get rid of this expected method, and replace it with whether a context manager, or having a callback. In both situations the wait will happen totally transparent to the consumer.

jrbenny35:
I personally like the idea of a context manager. I think that would help when we add different window types.

PR #69, opened 2017-02-04
Adding the initial representation of the tabbar and tab

jrbenny35:
@davehunt 

Figured I'd get started. I didn't do too much on the tabs as I'll leave that to another PR.


foxpuppet/windows/browser/tabbar.py
```diff
@@ -0,0 +1,46 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from selenium.webdriver.common.by import By
+
+from foxpuppet.windows.browser.tab import Tab
+
+
+class TabBar(Tab):
```
whimboo: The `TabBar` class should have nothing to do with the `Tab` class. It only contains `Tab` instances as properties. Also this is a `Region` too, right?

foxpuppet/windows/browser/tabbar.py
```diff
@@ -0,0 +1,46 @@
...
+
+    _new_tab_button = (By.ID, 'new-tab-button')
+    _tab_browser_locator = (By.ID, 'tabbrowser-tabs')
+    _tabs = (By.TAG_NAME, 'tab')
+
+    def __init__(self, selenium, *args, **kwargs):
+        self.selenium = selenium
+
+    @property
+    def all_tabs(self):
```
whimboo: s/all_tabs/tabs

foxpuppet/windows/browser/window.py
```diff
@@ -7,6 +7,7 @@
 
 from foxpuppet import expected
 from foxpuppet.windows import BaseWindow
+from foxpuppet.windows.browser.tabbar import TabBar
 from foxpuppet.windows.browser.notifications import BaseNotification
```
whimboo: nit: alphabetical order please.

foxpuppet/windows/browser/tabbar.py
```diff
@@ -0,0 +1,46 @@
...
+    """Representation of the tab bar which contains the tabs.
+
+    :param selenium: WebDriver Object
+    :type selenium:
+        :py:class:`~selenium.webdriver.remote.webdriver.WebDriver` object
+    """
+
+    _new_tab_button = (By.ID, 'new-tab-button')
+    _tab_browser_locator = (By.ID, 'tabbrowser-tabs')
+    _tabs = (By.TAG_NAME, 'tab')
```
whimboo: Generally a `_locator` suffix?

foxpuppet/windows/browser/window.py
```diff
@@ -32,6 +32,14 @@ def notification(self):
         except NoSuchElementException:
             return None  # no notification is displayed
 
+    @property
+    def tabs(self):
```
whimboo: Those are not the tabs, but the `tabBar`. As of now this would be misleading.

foxpuppet/windows/browser/tabbar.py
jrbenny35: So the nomenclature was tough as I thought ```self.browser.tabs.tabs``` was wrong. ```self.browser.tabs.all_tabs``` seemed better.

foxpuppet/windows/browser/window.py
jrbenny35: To the user, they don't need to access the tabbar directly, so I figured don't let them even know its there.

whimboo: There are a couple of other items in the tabbar. There are not only the tabs. Eg. pinned tabs, the arrows.

jrbenny35: Ah, good point. so ```self.browser.tabbar.tabs``` ?

jrbenny35:
I also think the context manager Henrik was mentioning in issue #65 could also apply to the opening of tabs.

foxpuppet/windows/browser/window.py
whimboo: Yes, I think so.

whimboo:
> I also think the context manager Henrik was mentioning in issue #65 could also apply to the opening of tabs.

Still something for a different PR, but yes it could also be used for opening a tab, whereby the tab implementation is different from the window one because of window_handles != chrome_window_handles.

foxpuppet/windows/base.py
```diff
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
+
```
davehunt: Nit: Remove this empty line.

foxpuppet/windows/browser/tab.py
```diff
@@ -0,0 +1,8 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+
+class Tab(object):
+    def __init__(self, selenium, *args, **kwargs):
+        self.selenium = selenium
```
davehunt: Yep, `Tab` should inherit from `Region`

davehunt: Also, this init has nothing custom, so it can be omitted. If there are likely to be a lot of regions within the tab bar then we might also want to consider making tab_bar a module and moving tab.py into it.

foxpuppet/windows/browser/tabbar.py
```diff
@@ -0,0 +1,46 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
davehunt: I'd prefer to name this `tab_bar`, personally.

foxpuppet/windows/browser/tabbar.py
```diff
@@ -0,0 +1,46 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from selenium.webdriver.common.by import By
+
+from foxpuppet.windows.browser.tab import Tab
+
+
+class TabBar(object):
```
davehunt: Inherit from `Region`

foxpuppet/windows/browser/tabbar.py
```diff
@@ -0,0 +1,46 @@
...
+    :type selenium:
+        :py:class:`~selenium.webdriver.remote.webdriver.WebDriver` object
+    """
+
+    _new_tab_button_locator = (By.ID, 'new-tab-button')
+    _tab_browser_locator = (By.ID, 'tabbrowser-tabs')
+    _tabs_locator = (By.TAG_NAME, 'tab')
+
+    def __init__(self, selenium, *args, **kwargs):
+        self.selenium = selenium
```
davehunt: No need for this.

foxpuppet/windows/browser/tabbar.py
```diff
@@ -0,0 +1,46 @@
...
+
+    """Representation of the tab bar which contains the tabs.
+
+    :param selenium: WebDriver Object
+    :type selenium:
+        :py:class:`~selenium.webdriver.remote.webdriver.WebDriver` object
+    """
+
+    _new_tab_button_locator = (By.ID, 'new-tab-button')
+    _tab_browser_locator = (By.ID, 'tabbrowser-tabs')
```
davehunt: This is unused. What is it for?

foxpuppet/windows/browser/tabbar.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def __init__(self, selenium, *args, **kwargs):
+        self.selenium = selenium
+
+    @property
+    def tabs(self):
+        """Returns a list of tabs.
+
+        :returns: :py:class:`~foxpuppet.window.browser.tab.Tab`
+        :return type: object
```
davehunt: The type is the class, given above. `object` is too vague.

foxpuppet/windows/browser/tabbar.py
```diff
@@ -0,0 +1,46 @@
...
+        """
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            return [Tab(self.selenium) for tabs in
+                    self.selenium.find_elements(*self._tabs_locator)]
```
davehunt: Each tab should have a root element.

foxpuppet/windows/browser/tabbar.py
```diff
@@ -0,0 +1,46 @@
...
+    def open_new_tab(self):
+        """Opens a new tab in the current window.
+
+        :returns: WebDriver Object
+        :return type:
+            :py:class:`~selenium.webdriver.remote.webdriver.WebDriver` object
+        """
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            return self.selenium.find_element(
+                *self._new_tab_button_locator).click()
```
davehunt: We should wait for the new tab to open.

foxpuppet/windows/browser/window.py
```diff
@@ -32,6 +32,14 @@ def notification(self):
         except NoSuchElementException:
             return None  # no notification is displayed
 
+    @property
+    def tabbar(self):
```
davehunt: Nit: `tab_bar`

foxpuppet/windows/browser/window.py
```diff
@@ -32,6 +32,14 @@ def notification(self):
...
+        """Creates the tab bar object.
+        :returns: :py:class:`~foxpuppet.window.browser.tabbar.TabBar`
+        :return type: object
```
davehunt: Please fix the return type.

foxpuppet/windows/browser/window.py
```diff
@@ -32,6 +32,14 @@ def notification(self):
...
+        """
+        return TabBar(self.selenium)
```
davehunt: We should pass a root element.

foxpuppet/windows/browser/tab.py
davehunt: I think we should at least implement a `close` method here.

foxpuppet/windows/browser/tabbar.py
jrbenny35: A tabs root element would be the tab itself?

davehunt: The root element would be where the definition of the tab starts, and all qualities of the tab would be descendant elements or attributes of that element. It is probably the `tab` element, but I've not looked at the document...

jrbenny35:
I am getting a weird error on the travis build. I don't get this locally, but when I did I switched selenium focus back to the window before checking how many tabs were currently open.

foxpuppet/windows/browser/tab.py
```diff
@@ -0,0 +1,39 @@
...
+    @property
+    def switch_to(self):
+        self.selenium.switch_to.window(self.selenium.current_window_handle)
+
+    def close(self):
+        from foxpuppet.windows.browser.tab_bar import TabBar
+        from foxpuppet.windows.browser.window import BrowserWindow
+
+        window = BrowserWindow(
+            self.selenium, self.selenium.current_window_handle)
```
whimboo: The tab should have a weak reference to the tabbrowser instance, so all this would not be necessary.

foxpuppet/windows/browser/tab.py
```diff
@@ -0,0 +1,39 @@
...
+
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            tabbar = TabBar(window, Tab)
+            current_tabs = tabbar.tabs
```
whimboo: Oh well, do not use `tabbar.tabs` for that but `window_handles`, which are the list of open tabs across all windows.

jrbenny35: Didn't know ```window_handles``` returned the tabs as well!

foxpuppet/windows/browser/tab.py
jrbenny35: I don't quite understand the ```Region``` part. What would be the correct way to get a ```Tab``` object?

foxpuppet/windows/browser/tab.py
whimboo: `window_handles` are tabs, and `chrome_window_handles` are the actual chrome windows in Marionette. Sounds confusing? It is... :)

foxpuppet/windows/browser/tab.py
whimboo: I think we should drop this for now given the discussion below.

jrbenny35: Dave wants to have a close button within the ```Tab``` class, so I think I'd still need to create an ```Tab``` object, unless I am not following what you're saying.

whimboo: Sorry, but I do not understand your last comment.

Why you would have to create a Tab instance when you already have one which in this case has the close method available?

jrbenny35: That's what I am saying. 

To create a ```Tab``` Instance I need to do what is on line 20/21, and pass that too Tab. At least that is how I am seeing it. I don't quite get the ```Region``` part so maybe that has me a bit confused.

whimboo: Line 20/21 of this file is nothing for me. Not sure what you are referring to. 

Also the `close` method is part of the Tab class. Why do you need a new instance?

foxpuppet/windows/browser/tab.py
davehunt: There's some confusion here. The tabs in the tab bar are not the window handles, but are associated with window handles. As FoxPuppet is for UI modeling, the tab refers to the visual artifact that the user can click/drag, and contains a button that can be clicked to close:

screenshot 2017-03-10 11 46 09

davehunt:
Does this have any motivation behind expanding the coverage of the Firefox UI? I think it helps to have a need to influence our design decisions. For example, I wrote the notifications handling because we want to be able to write an end-to-end extension installation test. Perhaps there are Firefox UI tests that will require us to model the tabs? If so, let's reference those and use them to help us to decide on a design for this API.

jrbenny35:
I have motivation as this is kinda my baby and I've been learning a ton from you and henrik. I guess there isn't a specific need for this at the moment however.

foxpuppet/windows/browser/tab.py
whimboo: I see. So you don't want to fallback to `window_handles` but do really everything via the UI elements? In such a case we might indeed need a weakref to the `tabtab` class.

davehunt:
> I have motivation as this is kinda my baby and I've been learning a ton from you and henrik. I guess there isn't a specific need for this at the moment however.

Well that's good motivation, too. I was just wondering if we have a consumer of the API to keep in mind. For example, a test that reorders the tabs using drop and drop, or pins a tab via a context-click.

whimboo:
> Well that's good motivation, too. I was just wondering if we have a consumer of the API to keep in mind. For example, a test that reorders the tabs using drop and drop, or pins a tab via a context-click.

There is nothing we even have in mind for that in firefox-ui tests. So maybe we should focus more on required APIs first? Not sure if there are already some from other teams.

jrbenny35:
I got some good experience with the ```Page``` ```Region``` model from moving some addons tests. I've updated this to work but I think the creation of the ```TabBar``` class could still use some help. I don't think it's wrong but it feels off. I'll probably play with it.

jrbenny35:
I am going to close this and update it. mozilla-ockbox is going to use Foxpuppet but we need control of the tabs since clicking on the extension opens a new tab.

PR #82, opened 2017-05-26
Fix breakage from Firefox 55 and coverage from Firefox ESR

davehunt:


davehunt:
Looks like coverage has been restored, and all tests are passing. Feel free to merge after reviewing if you're happy @jrbenny35.

jrbenny35:
I like this better than supporting the older 52. Thanks!

jrbenny35:
Could you add a note to the docs about the new support range?

davehunt:
> Could you add a note to the docs about the new support range?

The support range hasn't changed. We always planned ESR->Nightly.. All I've done is dropped 52 and earlier, which is before ESR.

PR #87, opened 2017-05-26
Update travis file to always download latest geckodriver

jrbenny35:


jrbenny35:
There seem to be random errors that occur. Sometimes the script that install geckodriver fails and it doesn't seem to really have a solid reason.

silne30:
LGTM bitmoji

PR #89, opened 2017-05-31
Try updating to trusty distribution

jrbenny35:
Seems the travis build can be flaky when trying to get the latest geckodriver. I don't really see a reason why it should however. This is just a test to see if its something simple.


silne30:
Can we run the build a few times in succession? Like repeat the build a few times to make sure this isn't a false positive.

jrbenny35:
> Can we run the build a few times in succession? Like repeat the build a few times to make sure this isn't a false positive.

Yes, maybe like 5 or so?

jrbenny35:
Hmm, still looks to be flaky. I may need to try and get the URL multiple times

jrbenny35:
Found something: 
```
GECKODRIVER_DOWNLOAD={
  "message": "API rate limit exceeded for 73.202.75.208. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)",
  "documentation_url": "https://developer.github.com/v3/#rate-limiting"
}
```
So looks like it works but when run on multiple PR's or builds it eventually hits this limiter.

jrbenny35:
@silne30 what do you think of the solution within ```geckodriver.py``` ? It will get the latest geckodriver each time but I needed to install ```requests``` and ```beautifulsoup4``` to do it. 

I could make this a fixture too.

silne30:
@jrbenny35 So what is the resolution here. Did we just wind up reverting the geckodriver? Do we still need a merge? There are a lot of commits to siphon through.

jrbenny35:
@silne30 I added the commands to ```geckodriver.py```. It downloads the latest version of geckodriver and then there are a few commands within travis to install it.

silne30:
LGTM

jrbenny35:
So the github api allows for only 50 requests an our from a certain IP. I think for now I will just update the geckodriver url manually.

Issue #72, opened 2017-02-22
Create Tracking Protection Shield object


rbillings:
When Tracking Protection is on, a shield displays in the toolbar.  An automated test needs to be written to verify the presence or absence of the shield icon.

whimboo:
There are actually several places where the shield appears. As you say it is visible on the left side of the location bar, but there are also settings inside the identity popup. Given that we have slow progress on all that, you might better want to write this as a firefox-ui test those days. Especially if a test is urgent.

jrbenny35:
@whimboo could you give me a link to where in marionette there is a test for this? Or maybe where it is used? I can't seem to find anything useful.

whimboo:
You will have to look at least for the identity popup related tests in the firefox-ui subfolder. Not sure what is implemented as of today.

jrbenny35:
@rbillings So I wrote something for this but there appears to be some difficulty determining if the shield is displayed or not. It looks like it is just hidden, and it doesn't have any css properties I can check for. Just wanted to give you an update.

davehunt:
@jrbenny35 this could be related to [bug 1094246](https://bugzilla.mozilla.org/show_bug.cgi?id=1094246).

jrbenny35:
Fixed in #92

PR #92, opened 2017-06-08
Add Tracking protection icon api and test

jrbenny35:
@davehunt @whimboo @rbillings 

Fix for #72 

Found a simple solution that I missed before!
I also am not 100% on the api for this so I am open to suggestions. I don't know what you would do with the icon besides check if it was there.

Also, I am not keen on navigating to cnn.com but I don't think we could mock a webserver that triggers firefox's tracking protection stuff.

### DIFF
--- a/foxpuppet/windows/browser/navbar.py
+++ b/foxpuppet/windows/browser/navbar.py
@@ -2,33 +2,28 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-
 from selenium.webdriver.common.by import By
-from selenium.webdriver.support.wait import WebDriverWait
-import time
+
+from foxpuppet.region import Region
 
 
-class NavBar(object):
+class NavBar(Region):
 
-    """Representation of the NavBar
-    :param selenium: WebDriver Object
-    :type selenium:
-        :py:class:`~selenium.webdriver.remote.webdriver.WebDriver` object
+    """Representation of the Navigation Bar
+    :param window: Window object this region appears in.
+    :param root: element that serves as the root for the region.
+    :type window: :py:class:`~.windows.BaseWindow`
+    :type root: :py:class:`~selenium.webdriver.remote.webelement.WebElement`
     """
-
-    _tracking_protection_shield_locator = (By.CSS_SELECTOR, '#tracking-protection-icon:-moz-lwtheme')
-
-    def __init__(self, selenium, *args, **kwargs):
-        self.selenium = selenium
+    _tracking_protection_shield_locator = (By.ID, 'tracking-protection-icon')
 
     @property
-    def tracking_shield(self):
-        """Returns the Tacking Protection enabled shield"""
-        time.sleep(30)
+    def is_tracking_shield_displayed(self):
+        """Tracking Protection shield
+        :returns: True or False if the Tracking Shield is displayed
+        :type return: boolean
+        """
         with self.selenium.context(self.selenium.CONTEXT_CHROME):
-            return self.selenium.find_element(
+            el = self.selenium.find_element(
                 *self._tracking_protection_shield_locator)
-            #WebDriverWait(self.selenium, timeout=10).until(
-            #    lambda _: self.selenium.find_element(
-            #        *self._tracking_protection_shield_locator)
-            #)
+            return bool(el.get_attribute('state'))
--- a/foxpuppet/windows/browser/navbar.py
+++ b/foxpuppet/windows/browser/navbar.py
@@ -1,0 +1,34 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support.wait import WebDriverWait
+import time
+
+
+class NavBar(object):
+
+    """Representation of the NavBar
+    :param selenium: WebDriver Object
+    :type selenium:
+        :py:class:`~selenium.webdriver.remote.webdriver.WebDriver` object
+    """
+
+    _tracking_protection_shield_locator = (By.CSS_SELECTOR, '#tracking-protection-icon:-moz-lwtheme')
+
+    def __init__(self, selenium, *args, **kwargs):
+        self.selenium = selenium
+
+    @property
+    def tracking_shield(self):
+        """Returns the Tacking Protection enabled shield"""
+        time.sleep(30)
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            return self.selenium.find_element(
+                *self._tracking_protection_shield_locator)
+            #WebDriverWait(self.selenium, timeout=10).until(
+            #    lambda _: self.selenium.find_element(
+            #        *self._tracking_protection_shield_locator)
+            #)
--- a/foxpuppet/windows/browser/window.py
+++ b/foxpuppet/windows/browser/window.py
@@ -7,6 +7,7 @@ from selenium.webdriver.common.by import By
 
 from foxpuppet import expected
 from foxpuppet.windows import BaseWindow
+from foxpuppet.windows.browser.navbar import NavBar
 from foxpuppet.windows.browser.notifications import BaseNotification
 
 
@@ -22,6 +23,10 @@ class BrowserWindow(BaseWindow):
     _notification_locator = (
         By.CSS_SELECTOR, '#notification-popup popupnotification')
     _tab_browser_locator = (By.ID, 'tabbrowser-tabs')
+
+    @property
+    def navbar(self):
+        return (NavBar(self.selenium))
 
     @property
     def notification(self):
--- a/foxpuppet/windows/browser/window.py
+++ b/foxpuppet/windows/browser/window.py
@@ -26,7 +26,14 @@ class BrowserWindow(BaseWindow):
 
     @property
     def navbar(self):
-        return (NavBar(self.selenium))
+        """Provides access to the Navigation Bar.
+        :returns: :py:class:`~foxpuppet.windows.browser.navbar.NavBar`
+        :return type: object
+        """
+        window = BaseWindow(self.selenium, self.selenium.current_window_handle)
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            el = self.selenium.find_element(*self._nav_bar_locator)
+            return NavBar(window, el)
 
     @property
     def notification(self):
--- a/requirements/tests.txt
+++ b/requirements/tests.txt
@@ -1,2 +1,2 @@
-pytest==3.1.1
-pytest-selenium==1.10.0
+pytest==3.1.3
+pytest-selenium==1.11.0
--- a/tests/test_browser_model.py
+++ b/tests/test_browser_model.py
@@ -2,6 +2,9 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import pytest
+from selenium.webdriver.support.wait import WebDriverWait
+
 
 def test_initial_browser_window(foxpuppet):
     """Tests initial state of browser windows"""
@@ -36,16 +39,16 @@ def test_close_window(foxpuppet):
 def test_switch_to(foxpuppet, selenium):
     """Test Switch to function"""
     foxpuppet.browser.open_window()
-
     # Switch to originally window opened by pytest
     foxpuppet.browser.switch_to()
     assert foxpuppet.browser.handle == selenium.current_window_handle
 
 
-def test_tracking_protection_sheild(foxpuppet, selenium):
-    browser = foxpuppet.browser.open_window(private=True)
-    browser.switch_to()
-    selenium.get('http://cnn.com')
-    assert browser.navbar.tracking_shield
-    selenium.get('http://support.mozilla.org')
-    assert foxpuppet.window_manager.windows[1].navbar.tracking_shield
+@pytest.mark.firefox_preferences({'privacy.trackingprotection.enabled': True})
+def test_tracking_protection_shield(foxpuppet, selenium):
+    """Tests if the tracking protection icon displays"""
+    browser = foxpuppet.browser
+    assert not browser.navbar.is_tracking_shield_displayed
+    selenium.get('https://www.washingtonpost.com/')
+    WebDriverWait(selenium, timeout=5).until(
+        lambda _: browser.navbar.is_tracking_shield_displayed)
--- a/tests/test_browser_model.py
+++ b/tests/test_browser_model.py
@@ -40,3 +40,12 @@ def test_switch_to(foxpuppet, selenium):
     # Switch to originally window opened by pytest
     foxpuppet.browser.switch_to()
     assert foxpuppet.browser.handle == selenium.current_window_handle
+
+
+def test_tracking_protection_sheild(foxpuppet, selenium):
+    browser = foxpuppet.browser.open_window(private=True)
+    browser.switch_to()
+    selenium.get('http://cnn.com')
+    assert browser.navbar.tracking_shield
+    selenium.get('http://support.mozilla.org')
+    assert foxpuppet.window_manager.windows[1].navbar.tracking_shield
--- a/foxpuppet/windows/base.py
+++ b/foxpuppet/windows/base.py
@@ -53,5 +53,4 @@ class BaseWindow(object):
 
     def switch_to(self):
         """Switches focus for Selenium commands to this window."""
-
         self.selenium.switch_to.window(self.handle)


foxpuppet/windows/browser/navbar.py
```diff
@@ -0,0 +1,32 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+
+from selenium.webdriver.common.by import By
+
+
+class NavBar(object):
```
whimboo: This should be based on `Region` or?

foxpuppet/windows/browser/window.py
```diff
@@ -24,6 +25,15 @@ class BrowserWindow(BaseWindow):
     _tab_browser_locator = (By.ID, 'tabbrowser-tabs')
 
     @property
+    def is_tracking_shield_displayed(self):
```
whimboo: The window needs a `navBar` property, which is an instance of the `NavBar` class. Why don't you want to setup a hierarchical structure?

tests/conftest.py
```diff
@@ -42,3 +42,9 @@ def browser(foxpuppet):
 @pytest.fixture
 def foxpuppet(selenium):
     return FoxPuppet(selenium)
+
+
+@pytest.fixture
+def capabilities(capabilities):
```
whimboo: Why does it need to be added now?

tests/test_browser_model.py
```diff
@@ -36,7 +36,15 @@ def test_close_window(foxpuppet):
 def test_switch_to(foxpuppet, selenium):
     """Test Switch to function"""
     foxpuppet.browser.open_window()
-
     # Switch to originally window opened by pytest
     foxpuppet.browser.switch_to()
     assert foxpuppet.browser.handle == selenium.current_window_handle
+
+
+def test_tracking_protection_sheild(foxpuppet, selenium):
```
whimboo: s/sheild/shield/

tests/test_browser_model.py
```diff
@@ -36,7 +36,15 @@ def test_close_window(foxpuppet):
...
+    """Tests if the tracking protection icon displays"""
+    browser = foxpuppet.browser.open_window(private=True)
+    browser.switch_to()
+    assert not browser.is_tracking_shield_displayed
```
whimboo: Is that the only method @rbillings needs? I don't think so. Please add everything what's necessary. Btw where are the requirements?

foxpuppet/windows/browser/window.py
jrbenny35: Heh, I had it like that at first.

tests/test_browser_model.py
jrbenny35: From what she has told me, this is all she needs to check.

jrbenny35: And what do you mean by requirements?

whimboo: @rbillings what are your requirements for the tracking shield support? What has to be covered?

rbillings:
I agree that it's not ideal to navigate to cnn.com, but it's a real world site with a value that is unlikely to change. R+

jrbenny35:
Looks like something changed with 56, but I have updated the addons name  location method and it seems to be working.

jrbenny35:
@davehunt @whimboo could one of you look at this when you guys have a chance?

davehunt:
Sorry for the delay @jrbenny35. I'll try to get to this today.

foxpuppet/windows/browser/navbar.py
```diff
@@ -0,0 +1,39 @@
...
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from selenium.webdriver.common.by import By
+
+from foxpuppet.region import Region
+
+
+class NavBar(Region):
+
+    """Representation of the NavBar
```
davehunt: I'd refer to this expanded as 'navigation bar' in the documentation.

foxpuppet/windows/browser/navbar.py
```diff
@@ -0,0 +1,39 @@
...
+    :param selenium: WebDriver Object
```
davehunt: A `Region` object expects a `Window` object as the first argument. Note that this is different from the implementation in PyPOM, which is for web pages/regions.

foxpuppet/windows/browser/navbar.py
```diff
@@ -0,0 +1,39 @@
...
+    :param root: Root WebElement
+    :type selenium:
+        :py:class:`~selenium.webdriver.remote.webdriver.WebDriver` object
+    :type root:
+        :py:class:`~selenium.webdriver.remote.webdriver.WebDriver` object
```
davehunt: This is a `WebElement` as documented [here](https://github.com/mozilla/FoxPuppet/blob/master/foxpuppet/region.py).

foxpuppet/windows/browser/navbar.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+
+    _root_locator = (By.ID, 'nav-bar')
```
davehunt: This isn't used anywhere. Note that our implementation of `Region` differs from PyPOM, and this locator is not automatically used for looking up root elements. We could port this feature to our implementation though.

foxpuppet/windows/browser/navbar.py
```diff
@@ -0,0 +1,39 @@
...
+    _tracking_protection_shield_locator = (By.ID, 'tracking-protection-icon')
+
+    def __init__(self, selenium, *args, **kwargs):
+        self.selenium = selenium
```
davehunt: It should not be necessary to override this method from the parent class.

foxpuppet/windows/browser/navbar.py
```diff
@@ -0,0 +1,39 @@
...
+
+    @property
+    def is_tracking_shield_displayed(self):
+        """Tracking Protection shield
+        :returns:
+            :py:class:`~selenium.webdriver.remote.webdriver.WebDriver` object
+        :return type: object
```
davehunt: We should note in the docstring that the property returns a boolean indicating if the tracking shield is displayed.

foxpuppet/windows/browser/navbar.py
```diff
@@ -0,0 +1,39 @@
...
+        """
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            if self.selenium.find_element(
+                *self._tracking_protection_shield_locator). \
+                    get_attribute('state'):
+                return True
+            return False
```
davehunt: You can avoid the logic by doing something like:

```python
el = self.root.find_element(*self._tracking_protection_shield_locator)
return bool(el.get_attribute('state'))
```

Note the use of `self.root`, which will locate the element within the root element rather than in the entire document. In your case this shouldn't matter as the element has an ID, but it's a good practice to restrict element lookup within regions.

foxpuppet/windows/browser/notifications/addons.py
```diff
@@ -33,8 +33,8 @@ def addon_name(self):
         """
         with self.selenium.context(self.selenium.CONTEXT_CHROME):
             if self.window.firefox_version >= 55:
-                return self.root.find_anonymous_element_by_attribute(
-                    'anonid', 'description').find_element(
+                return self.root.find_element(
+                    By.ID, 'addon-webext-perm-header').find_element(
```
davehunt: Can you take care of this in a separate pull request? It's not related to the navigation bar or tracking shield.

foxpuppet/windows/browser/window.py
```diff
@@ -24,6 +25,14 @@ class BrowserWindow(BaseWindow):
     _tab_browser_locator = (By.ID, 'tabbrowser-tabs')
 
     @property
+    def navbar(self):
+        """Provides access to the navbar.
```
davehunt: Use 'navigation bar' in docs.

foxpuppet/windows/browser/window.py
```diff
@@ -24,6 +25,14 @@ class BrowserWindow(BaseWindow):
...
+        :returns: :py:class:`~foxpuppet.windows.browser.navbar.NavBar`
+        :return type: object
+        """
+        return NavBar(self.selenium)
```
davehunt: We need to pass the window and root element, unless we implement use of `_root_locator` like we have in PyPOM.

tests/test_browser_model.py
```diff
@@ -36,7 +36,15 @@ def test_close_window(foxpuppet):
...
     foxpuppet.browser.open_window()
-
     # Switch to originally window opened by pytest
     foxpuppet.browser.switch_to()
     assert foxpuppet.browser.handle == selenium.current_window_handle
+
+
+def test_tracking_protection_shield(foxpuppet, selenium):
+    """Tests if the tracking protection icon displays"""
+    browser = foxpuppet.browser.open_window(private=True)
```
davehunt: Instead of using a private window, you can just set preference "privacy.trackingprotection.enabled" to True.

tests/test_browser_model.py
```diff
@@ -36,7 +36,15 @@ def test_close_window(foxpuppet):
...
+    browser.switch_to()
+    assert not browser.navbar.is_tracking_shield_displayed
+    selenium.get('http://cnn.com')
```
davehunt: I'd like to avoid using an external website. We already have a local http server for another test, could you see if we can create some test content that triggers the tracking protection?

foxpuppet/windows/browser/navbar.py
jrbenny35: When I see region I think PyPOM 🤣

foxpuppet/windows/browser/navbar.py
jrbenny35: So regular ```self.find_element()``` ?

tests/test_browser_model.py
jrbenny35: According to: https://developer.mozilla.org/en-US/Firefox/Privacy/Tracking_Protection it is just a list of URL's.

foxpuppet/windows/browser/navbar.py
whimboo: And to add it's also faster when you search only for elements starting by a sub node.

Otherwise I wonder why we need this method at all? Why can't a test not use the tracking protection shield property to retrieve the element, and then retrieve the attribute itself? I don't see a benefit here in wrapping that.

tests/test_browser_model.py
whimboo: I'm fairly sure that we could use `execute_script` in chrome scope to add locally served pages.

foxpuppet/windows/browser/navbar.py
davehunt: > So regular `self.find_element()` ?

No, that's PyPOM. We could add a `find_element` method to our `Region` though to introduce a similar behaviour. Otherwise you'd want `self.root.find_element`.

> Otherwise I wonder why we need this method at all? Why can't a test not use the tracking protection shield property to retrieve the element, and then retrieve the attribute itself? I don't see a benefit here in wrapping that.

The concept of page objects (which this is based on) means the consumer doesn't need to understand the API of Selenium. It also means that theoretically Selenium could be substituted for another tool.

whimboo: Sure, but you really want a method for each needed property/attribute? Wow!

davehunt: > Sure, but you really want a method for each needed property/attribute? Wow!

No, this is being used as an indication that the icon is displayed. If it wasn't for the bug related to `is_displayed` for chrome elements, we'd likely be using that.

tests/test_browser_model.py
jrbenny35: @whimboo me and dave found this: https://dxr.mozilla.org/mozilla-central/source/toolkit/components/url-classifier/tests/mochitest/test_classifier.html

I am thinking we just add a localhost URL to it, since we have a webserver. 

Do you know if there are any examples of this?

whimboo: Hm, looks a  it too complex to do. see the referenced .js file for `classifierHelper´ in this test.

jrbenny35: Which database would I need to add the review too?

whimboo: Ups, something scrambled my comment yesterday. so what I wanted to say is that this looks to be too complex to do right now. Maybe we should go ahead with a remote site. Maybe use one which is less annoying than CNN.

We just got a new mixed content setup (bug 1351694) and it looks like we could use one of those pages.

jrbenny35: I think a site like cnn would be fine as users may run these tests locally, and I'd rather it be a site most people would know.

whimboo: This website takes x-times longer to load.

jrbenny35:
Found a decent URL that doesn't take too long to load. The tests are failing but I have a fix in #96

foxpuppet/windows/browser/window.py
```diff
@@ -24,6 +25,15 @@ class BrowserWindow(BaseWindow):
     _tab_browser_locator = (By.ID, 'tabbrowser-tabs')
 
     @property
+    def navbar(self):
+        """Provides access to the Navigation Bar.
+        :returns: :py:class:`~foxpuppet.windows.browser.navbar.NavBar`
+        :return type: object
+        """
+        window = BaseWindow(self.selenium, self.selenium.current_window_handle)
+        return NavBar(window, self.document_element)
```
davehunt: The second argument should be the navbar element.

tests/test_browser_model.py
```diff
@@ -1,6 +1,13 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
+import pytest
+
+
+@pytest.fixture
+def firefox_options(firefox_options):
+    firefox_options.set_preference('privacy.trackingprotection.enabled', True)
```
davehunt: Since pytest-selenium 0.11.0 you should be able to set this on tests that require it using:

```python
@pytest.mark.firefox_preferences({'privacy.trackingprotection.enabled': True})
```

foxpuppet/windows/browser/window.py
jrbenny35: But I would have to find that element correct?

tests/test_browser_model.py
jrbenny35: 😍 that is just beautiful

jrbenny35:
@davehunt any idea how to make the wait for the shield a bit more robust? It seems to load sometimes before the page is done and sometimes it loads after the page has finished loading.

whimboo:
You should use a wait in the test anyway, right? You don't know how long it takes until the page has been finished loading and the icon is being displayed.

jrbenny35:
> You should use a wait in the test anyway, right? You don't know how long it takes until the page has been finished loading and the icon is being displayed.

I don't want to use an implicit wait I am just waiting for the page to finish loading. That only works with 55 and up though...

whimboo:
Sorry, but I don't understand what you are trying to achieve here. Maybe you can tell it to me in person today.

jrbenny35:
This is failing due to #95

whimboo:
> Also sometimes the page doesn't load. It seems washington post has some pretty insane adds recently. I may have to try and find another URL

Why? I already gave you a working and fast loading URL.

tests/test_browser_model.py
```diff
@@ -36,7 +38,16 @@ def test_close_window(foxpuppet):
...
     assert foxpuppet.browser.handle == selenium.current_window_handle
+
+
+@pytest.mark.firefox_preferences({'privacy.trackingprotection.enabled': True})
+def test_tracking_protection_shield(foxpuppet, selenium):
+    """Tests if the tracking protection icon displays"""
+    browser = foxpuppet.browser
+    assert not browser.navbar.is_tracking_shield_displayed
+    selenium.get('https://www.washingtonpost.com/')
+    assert WebDriverWait(selenium, timeout=30).until(
```
whimboo: This should never take more than 5s! Otherwise something is really wrong.

jrbenny35: Yeah I agree. I think its the website. Like you were saying, running on AWS there is added latency. 

Also what URL did you give me? I don't remember which one you gave me I apologize.

whimboo: No. As explained today Marionette waits until the page has been loaded. So the call to `get` will not return earlier. Showing the icon is usually done immediately.

Regarding the webpage check the former most likely now outdated review comments.

jrbenny35: Gotcha. Looking at the results in travis, sometimes the ```get``` will timeout. That's what I was originally mentioning about trying out a different URL.

jrbenny35:
I need #95 merged before this test can work properly.

davehunt:
> I need #95 merged before this test can work properly.

You can always cherry-pick the commit from #95 to this pull request. Then if #95 is merged you can rebase, or if this lands first we can close #95 without merging.

foxpuppet/windows/browser/window.py
```diff
@@ -24,6 +25,17 @@ class BrowserWindow(BaseWindow):
...
     @property
+    def navbar(self):
+        """Provides access to the Navigation Bar.
+        :returns: :py:class:`~foxpuppet.windows.browser.navbar.NavBar`
+        :return type: object
+        """
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            navbar = self.selenium.find_element(*self._nav_bar_locator)
+        window = BaseWindow(self.selenium, self.selenium.current_window_handle)
+        return NavBar(window, navbar)
```
davehunt: I suspect it makes no difference, but I think this would read better as:

```python
window = BaseWindow(self.selenium, self.selenium.current_window_handle)
    el = self.selenium.find_element(*self._nav_bar_locator)
    return NavBar(window, el)
```

What do you think?

tests/test_browser_model.py
```diff
@@ -1,6 +1,8 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
+import pytest
```
davehunt: Nit: Add a blank line between license and imports.

tests/test_browser_model.py
```diff
@@ -36,7 +38,16 @@ def test_close_window(foxpuppet):
...
+
+
+@pytest.mark.firefox_preferences({'privacy.trackingprotection.enabled': True})
+def test_tracking_protection_shield(foxpuppet, selenium):
+    """Tests if the tracking protection icon displays"""
+    browser = foxpuppet.browser
+    assert not browser.navbar.is_tracking_shield_displayed
+    selenium.get('https://www.washingtonpost.com/')
+    assert WebDriverWait(selenium, timeout=5).until(
+        lambda _: browser.navbar.is_tracking_shield_displayed)
```
davehunt: Are you seeing intermittent failures when the tracking icon is not displayed? Generally I'd prefer not to have waits in the tests, but I think this makes the most sense at least for now. Note that you don't need the `assert` as the wait will either result in `True` or will raise a Timeout exception.

jrbenny35:
> You can always cherry-pick the commit from #95 to this pull request. Then if #95 is merged you can rebase, or if this lands first we can close #95 without merging.

Well I don't think travis will build with merge conflicts

davehunt:
> Well I don't think travis will build with merge conflicts

Probably not, but I don't understand why that would prevent you from cherry-picking the commit? You should be able to resolve the conflicts locally and force-push to the branch.

davehunt:
@jrbenny35 please resolve the conflicts and push an update. The master branch is now in a good state and is passing on Travis CI.

jrbenny35:
I forgot to rebase before the push...I think the update branch command should work...

requirements/tests.txt
```diff
@@ -1,2 +1,2 @@
-pytest==3.1.1
-pytest-selenium==1.10.0
+pytest==3.1.3
+pytest-selenium==1.11.0
```
davehunt: These shouldn't be needed now?

jrbenny35: Ahh....yeah I think that's from the last time.

PR #94, opened 2017-06-10
Update geckodriver

jrbenny35:
As per title.


silne30:
@jrbenny35 Good to go!

jrbenny35:
Trusty is very much still a beta heh

PR #96, opened 2017-06-22
Change addon name locator

jrbenny35:
This should fix the fails on nightly.

### DIFF
--- a/foxpuppet/windows/browser/notifications/addons.py
+++ b/foxpuppet/windows/browser/notifications/addons.py
@@ -20,8 +20,6 @@ class AddOnInstallBlocked(BaseNotification):
 class AddOnInstallConfirmation(BaseNotification):
     """Add-on install confirmation notification."""
 
-    _addon_name_locator = (
-        By.CSS_SELECTOR, '#addon-install-confirmation-content label')
     _web_ext_header_locator = (By.ID, 'addon-webext-perm-header')
     _web_ext_name_locator = (By.CSS_SELECTOR, '.addon-webext-name')
     _cancel_locator = (By.ID, 'addon-install-confirmation-cancel')
@@ -35,11 +33,15 @@ class AddOnInstallConfirmation(BaseNotification):
         """
         with self.selenium.context(self.selenium.CONTEXT_CHROME):
             if self.window.firefox_version >= 55:
-                return self.root.find_element(
-                    *self._web_ext_header_locator).find_element(
-                        *self._web_ext_name_locator).text
+                _addon_name_locator = (
+                    By.CSS_SELECTOR, '#addon-webext-perm-header > \
+                                      .addon-webext-name')
+                return self.root.find_element(*_addon_name_locator).text
             else:
-                label = self.root.find_element(*self._addon_name_locator)
+                _addon_name_locator = (
+                    By.CSS_SELECTOR, '#addon-install-confirmation-content \
+                                      label')
+                label = self.root.find_element(*_addon_name_locator)
                 return label.get_property('value')
 
     def cancel(self):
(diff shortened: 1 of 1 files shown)


foxpuppet/windows/browser/notifications/addons.py
```diff
@@ -33,8 +33,8 @@ def addon_name(self):
         """
         with self.selenium.context(self.selenium.CONTEXT_CHROME):
             if self.window.firefox_version >= 55:
-                return self.root.find_anonymous_element_by_attribute(
-                    'anonid', 'description').find_element(
+                return self.root.find_element(
+                    By.ID, 'addon-webext-perm-header').find_element(
```
davehunt: If we're not using anonymous nodes, then locators should be defined as class variables. As we're effectively replacing `_addon_name_locator` I think we can just replace this class variable within this if block.

foxpuppet/windows/browser/notifications/addons.py
```diff
@@ -33,9 +35,9 @@ def addon_name(self):
         """
         with self.selenium.context(self.selenium.CONTEXT_CHROME):
             if self.window.firefox_version >= 55:
-                return self.root.find_anonymous_element_by_attribute(
-                    'anonid', 'description').find_element(
-                        By.CSS_SELECTOR, '.addon-webext-name').text
+                return self.root.find_element(
+                    *self._web_ext_header_locator).find_element(
+                        *self._web_ext_name_locator).text
```
davehunt: Are you not able to reference this element directly using `#addon-webext-perm-header .addon-webext-name`? The only reason an intermediary node was used before was due to the parent node being anonymous.

foxpuppet/windows/browser/notifications/addons.py
```diff
@@ -22,6 +22,8 @@ class AddOnInstallConfirmation(BaseNotification):
 
     _addon_name_locator = (
         By.CSS_SELECTOR, '#addon-install-confirmation-content label')
+    _web_ext_header_locator = (By.ID, 'addon-webext-perm-header')
+    _web_ext_name_locator = (By.CSS_SELECTOR, '.addon-webext-name')
```
davehunt: I think we should just reassign the value of `_addon_name_locator` within the conditional logic. Once all supported Firefox versions use the same locator we just remove the conditional block and revert to a single value for the locator.

jrbenny35:
I need #95 merged before this test can work properly.

foxpuppet/windows/browser/notifications/addons.py
```diff
@@ -33,11 +31,15 @@ def addon_name(self):
         """
         with self.selenium.context(self.selenium.CONTEXT_CHROME):
             if self.window.firefox_version >= 55:
-                return self.root.find_anonymous_element_by_attribute(
-                    'anonid', 'description').find_element(
-                        By.CSS_SELECTOR, '.addon-webext-name').text
+                _addon_name_locator = (
+                    By.CSS_SELECTOR, '#addon-webext-perm-header > \
+                                      .addon-webext-name')
```
davehunt: One of these locators should still be on line 22. I would suggest this one, as it should ultimately become the only locator once the changes ride the trains to release and eventually ESR.

foxpuppet/windows/browser/notifications/addons.py
```diff
@@ -33,11 +31,15 @@ def addon_name(self):
...
+                return self.root.find_element(*_addon_name_locator).text
             else:
-                label = self.root.find_element(*self._addon_name_locator)
+                _addon_name_locator = (
+                    By.CSS_SELECTOR, '#addon-install-confirmation-content \
+                                      label')
```
davehunt: Can we avoid splitting the string over multiple lines? If this doesn't fit in 80 characters, split between arguments such as:

```python
_addon_name_locator = (
    By.CSS_SELECTOR,
    '#addon-install-confirmation-content label')
```

davehunt:
Are the Travis CI failures due to [bug 1376586](https://bugzilla.mozilla.org/show_bug.cgi?id=1376586) or something else?

jrbenny35:
> Are the Travis CI failures due to bug 1376586 or something else?

Yes

PR #104, opened 2017-07-20
Updated all documentation to comply with PEP257/258. Added flake8 pac…

jrbenny35:
Updates to Documentation and added flake8 extensions/packages to keep consistency.

### DIFF
--- a/foxpuppet/expected.py
+++ b/foxpuppet/expected.py
@@ -10,7 +10,7 @@ class new_browser_window_is_opened(object):
     """An expectation for checking that a new window is found.
 
     Returns:
-        A BrowserWindow object.
+        :py:class:`BrowserWindow`: Browser window.
 
     """
 
@@ -31,7 +31,7 @@ class new_browser_window_is_opened(object):
         """Check to see if a new window has opened.
 
         Returns:
-            obj: BrowserWindow object of new window.
+            :py:class:`BrowserWindow`: Opened window.
 
         """
         handles = list(set(self.selenium.window_handles) - set(self.handles))
--- a/foxpuppet/expected.py
+++ b/foxpuppet/expected.py
@@ -1,23 +1,39 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
+"""Module contained custom expected waits and conditions."""
 
 from __future__ import absolute_import
 
 
 class new_browser_window_is_opened(object):
-    """
-    An expectation for checking that a new window is found after requesting
-    it to be opened
+    """An expectation for checking that a new window is found.
 
-    :returns: A BrowserWindow Object
+    Returns:
+        A BrowserWindow object.
+
     """
 
     def __init__(self, selenium, handles):
+        """Create new_browser_window_is_opened object.
+
+        Args:
+            selenium:
+                (:py:class:`~selenium.webdriver.remote.webdriver.WebDriver`):
+                Firefox WebDriver object.
+            handles: (:obj:`list` of str): List of current Firefox window
+                handles.
+        """
         self.selenium = selenium
         self.handles = handles
 
     def __call__(self, *args, **kwargs):
+        """Check to see if a new window has opened.
+
+        Returns:
+            obj: BrowserWindow object of new window.
+
+        """
         handles = list(set(self.selenium.window_handles) - set(self.handles))
         if len(handles) == 1:
             from foxpuppet.windows import BrowserWindow
--- a/foxpuppet/foxpuppet.py
+++ b/foxpuppet/foxpuppet.py
@@ -1,6 +1,7 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
+"""FoxPuppet object for browser interaction."""
 
 from __future__ import absolute_import
 
@@ -8,16 +9,22 @@ from foxpuppet.windows import WindowManager
 
 
 class FoxPuppet(object):
-    """
-        Class that sets up the interface for interacting with the Firefox
-        browser.
+    """Set up the interface for interacting with the Firefox browser.
 
-        :param selenium: WebDriver object
-        :type selenium:
-            :py:class:`~selenium.webdriver.remote.webdriver.WebDriver`
+    Args:
+        selenium:
+            (:py:class:`~selenium.webdriver.remote.webdriver.WebDriver`):
+            Firefox WebDriver object.
     """
 
     def __init__(self, selenium):
+        """Create FoxPuppet object.
+
+        Args:
+            selenium:
+                (:py:class:`~selenium.webdriver.remote.webdriver.WebDriver`):
+                Firefox WebDriver object.
+        """
         self.selenium = selenium
         self.window_manager = WindowManager(selenium)
         # Need to ensure the first window is a browser window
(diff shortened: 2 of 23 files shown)


foxpuppet/windows/base.py
```diff
@@ -1,56 +1,68 @@
...
-        """ Returns the inner DOM window element.
+        """Return the inner DOM window element.
 
         :return: DOM window element.
         :return type:
             :py:class:`~selenium.webdriver.remote.webdriver.WebDriver` locator
-        """
 
+        Returns:
+            obj: Webdriver locator object for the DOM window element.
```
davehunt: Isn't this duplication of the preceding lines? Can we combine them?

foxpuppet/windows/base.py
```diff
@@ -1,56 +1,68 @@
...
     @property
     def firefox_version(self):
-        """ Major version of Firefox in use.
+        """Major version of Firefox in use.
 
         :returns: Major component of the Firefox version.
         :rtype: int
+
+        Returns:
+            int: Major component of the Firefox version.
```
davehunt: Again, can we avoid the duplication?

foxpuppet/windows/browser/navbar.py
```diff
@@ -1,27 +1,39 @@
...
     @property
     def is_tracking_shield_displayed(self):
-        """Tracking Protection shield
+        """Tracking Protection shield.
+
         :returns: True or False if the Tracking Shield is displayed
         :type return: boolean
+
+        Returns:
+            bool: True or False if the Tracking Shield is displayed.
```
davehunt: Last comment on the duplication, but if we can fix it we should fix it in all places.

foxpuppet/windows/browser/notifications/addons.py
```diff
@@ -1,17 +1,18 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
-
-from foxpuppet.windows.browser.notifications import BaseNotification
+"""Contains all current install notifications firefox will display."""
```
davehunt: nit: Firefox

foxpuppet/windows/browser/notifications/base.py
```diff
@@ -25,6 +45,10 @@ def label(self):
         """Provide access to the notification label.
 
         :returns: The notification label.
+
+        Returns:
+            str: The notification lable
```
davehunt: typo

foxpuppet/windows/base.py
jrbenny35: Will sphinx understand lines 54-55? According to the docs it must be how liners 51-52 are. Maybe I am missing something.

foxpuppet/windows/base.py
jrbenny35: This is what it looks like when I run python's ```help()```:

```document_element
     |      Return the inner DOM window element.
     |
     |      :return: DOM window element.
     |      :return type:
     |          :py:class:`~selenium.webdriver.remote.webdriver.WebDriver` locator
     |
     |      Returns:
     |          obj: Webdriver locator object for the DOM window element.
```

foxpuppet/windows/base.py
jrbenny35: This is what it looks like when I run python's ```help```:

```
firefox_version
     |      Major version of Firefox in use.
     |
     |      :returns: Major component of the Firefox version.
     |      :rtype: int
     |
     |      Returns:
     |          int: Major component of the Firefox version.
```

foxpuppet/foxpuppet.py
```diff
@@ -1,23 +1,26 @@
...
-    """
-        Class that sets up the interface for interacting with the Firefox
-        browser.
+    """Set up the interface for interacting with the Firefox browser.
 
-        :param selenium: WebDriver object
-        :type selenium:
-            :py:class:`~selenium.webdriver.remote.webdriver.WebDriver`
+    Args:
+        selenium: (:obj:`WebDriver`): Firefox WebDriver object.
```
davehunt: This drops the link to the Selenium docs. Could you restore this by using:

```
selenium: (:py:class:`~selenium.webdriver.remote.webdriver.WebDriver`): Firefox WebDriver object.
```

This comment should be applied to all arg and return types.

foxpuppet/windows/base.py
```diff
@@ -1,56 +1,54 @@
...
     def document_element(self):
-        """ Returns the inner DOM window element.
+        """Return the inner DOM window element.
 
-        :return: DOM window element.
-        :return type:
-            :py:class:`~selenium.webdriver.remote.webdriver.WebDriver` locator
-        """
+        Returns:
+            obj: WebDriver locator object for the DOM window element.
```
davehunt: This return the element, not the locator.

setup.cfg
```diff
@@ -1,2 +1,7 @@
 [coverage:run]
 source = foxpuppet
+
+[isort]
+line_length=79
```
davehunt: Isn't the default 80? Why do we need 79?

setup.cfg
```diff
@@ -1,2 +1,7 @@
...
+known_first_party = foxpuppet
+skip=tests
```
davehunt: Why are we skipping the tests?

tests/test_notifications.py
```diff
@@ -1,19 +1,20 @@
...
 from selenium.common.exceptions import TimeoutException
 
 from foxpuppet.windows.browser.notifications import BaseNotification
-from foxpuppet.windows.browser.notifications.addons import (
-    AddOnInstallBlocked,
-    AddOnInstallComplete,
-    AddOnInstallConfirmation)
+from foxpuppet.windows.browser.notifications.addons import (AddOnInstallBlocked, # noqa
+                                                            AddOnInstallComplete, # noqa
+                                                            AddOnInstallConfirmation) # noqa
```
davehunt: What rules were these lines breaking?

tests/webserver.py
```diff
@@ -1,6 +1,7 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
+"""Webserver used to server a webextion.xpi file for testing."""
```
davehunt: This class knows nothing about the web extension, it's just a simple webserver.

tests/test_notifications.py
jrbenny35: Well they are too long, and isort doesn't allow it to look any other way. Its kinda weird

davehunt: I think we should skip the isort warning, rather than all warnings. I'll take care of this in a followup.

davehunt:
@jrbenny35 I've made a few further enhancements, and rebased. This means I've force pushed to your branch, so you'll need to run `git pull --rebase` from your branch to get the latest changes. That's only if you want to continue working on this branch. If you're happy with my changes let me know and I'll clean up the commit history and merge to master.

foxpuppet/expected.py
```diff
@@ -10,7 +10,7 @@ class new_browser_window_is_opened(object):
     """An expectation for checking that a new window is found.
 
     Returns:
-        A BrowserWindow object.
+        :py:class:`BrowserWindow`: Browser window.
```
jrbenny35: I still don't like how this looks. I know we want to link the docs together but maybe there is a better way.

davehunt: Do you mean how it looks in the file, or in the rendered documentation?

jrbenny35: Yeah how it would be rendered to the user. But I guess they would understand it or if not the docs online would be good. Or maybe we could have a doc build step so they could have local access to them if they have a browser.

jrbenny35:
I think this is fine to merge.

davehunt:
> I think this is fine to merge.

Thanks, I'll clean it up, and rebase. We can follow up with further documentation improvements for readability (and I think we can generate documentation for more classes now).

PR #105, opened 2017-07-24
Add notification complete functions

jrbenny35:
This adds the ability to close the ```AddOnInstallComplete``` notification. I also added a test for it. I think we should wait until the docs PR is closed, as this does not include proper docstrings.


jrbenny35:
Ugh, looks like some old files got commited.

silne30:
Test ran correctly. LGTM.

davehunt:
I don't think this is needed, as the base notification's `close` works fine for me. I'm going to revert part of this change to restore Firefox ESR support in #110, and if it's not needed I think it makes sense to remove the test, too. I don't think we need to be exhaustive in making sure we can close all of the different notifications, unless there are genuine reasons for overriding the base methods.

Of course, if you have a need for this we can restore it, but as far as I can tell it's unnecessary.

jrbenny35:
> I don't think this is needed, as the base notification's close works fine for me. I'm going to revert part of this change to restore Firefox ESR support in #110, and if it's not needed I think it makes sense to remove the test, too. I don't think we need to be exhaustive in making sure we can close all of the different notifications, unless there are genuine reasons for overriding the base methods.
Of course, if you have a need for this we can restore it, but as far as I can tell it's unnecessary.

Ah you are right. I think I was running into another issue and this seemed to solve it.

Issue #108, opened 2017-08-01
No PyPi Deployment from TravisCI


jrbenny35:
We do not have a deploy step currently within the ```travis.yml``` file.

jrbenny35:
@davehunt the Foxpuppet pypi package is under your name so I guess you will have to provide the credentials to deploy it.

jrbenny35:
Fixed in #110

PR #110, opened 2017-08-04
Use build stages in Travis CI

davehunt:
Introduce build stages, including deploying to PyPI when the tests pass and a tag is pushed. This also fixes the ESR tests, which were incorrectly using release.

### DIFF
--- a/foxpuppet/windows/browser/notifications/addons.py
+++ b/foxpuppet/windows/browser/notifications/addons.py
@@ -44,25 +44,26 @@ class AddOnInstallConfirmation(BaseNotification):
     def cancel(self):
         """Cancel add-on install."""
         with self.selenium.context(self.selenium.CONTEXT_CHROME):
-            self.root.find_anonymous_element_by_attribute(
-                'anonid', 'secondarybutton').click()
+            if self.window.firefox_version >= 53:
+                # Notifications were restyled in Firefox 53
+                self.root.find_anonymous_element_by_attribute(
+                    'anonid', 'secondarybutton').click()
+            else:
+                self.root.find_element(*self._cancel_locator).click()
 
     def install(self):
         """Confirm add-on install."""
         with self.selenium.context(self.selenium.CONTEXT_CHROME):
-            self.root.find_anonymous_element_by_attribute(
-                'anonid', 'button').click()
+            if self.window.firefox_version >= 53:
+                # Notifications were restyled in Firefox 53
+                self.root.find_anonymous_element_by_attribute(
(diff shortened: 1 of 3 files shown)


davehunt:
This *should* fix #108 by doing a deploy only when all tests pass. I also reintroduced tests for ESR, which I mistakenly removed in an earlier commit and had to include a couple of fixes. We'll only know if a deploy works when we push a tag, but I'm not overly confident. Usually when a deploy condition is not met there is a message to indicate why a deployment doesn't happen. I don't see this, but maybe it's a quirk of build stages. I've asked about it in #travis on IRC.

foxpuppet/windows/browser/notifications/addons.py
```diff
@@ -44,26 +44,27 @@ def addon_name(self):
...
+                self.root.find_anonymous_element_by_attribute(
+                    'anonid', 'button').click()
+            else:
+                self.root.find_element(*self._confirm_locator).click()
 
 
 class AddOnInstallComplete(BaseNotification):
     """Add-on install complete notification."""
 
-    def close(self):
```
jrbenny35: Can you explain why you removed this?

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{27,36}, flake8
+envlist = py{27,36}, flake8, docs
```
jrbenny35: How were the docs being built before?

davehunt: Good question, possibly tox ignores a missing environment in the list if there's a section for it.

jrbenny35:
Looks good!

jrbenny35:
> This should fix #108 by doing a deploy only when all tests pass. I also reintroduced tests for ESR, which I mistakenly removed in an earlier commit and had to include a couple of fixes. We'll only know if a deploy works when we push a tag, but I'm not overly confident. Usually when a deploy condition is not met there is a message to indicate why a deployment doesn't happen. I don't see this, but maybe it's a quirk of build stages. I've asked about it in #travis on IRC.

Was there any response on this?

davehunt:
> Was there any response on this?

Unfortunately not. I'm going to go ahead and merge, if we have issues with the deploy we can raise an issue. The build stages feature is in beta, but I wouldn't want to do an automated deploy without it as it ensures all tests are passing rather than just one.

davehunt:
"Skipping a deployment with the pypi provider because this is not a tagged commit" sounds good. Feel free to push a tag @jrbenny35 to see if we get a deployment! :)

PR #113, opened 2017-08-10
Enable headless Firefox for beta and developer edition

davehunt:


davehunt:
Looks like Firefox 56 has some issues with notification visibility. Let's close this and retarget Firefox 57, as Nightly is doing well with headless mode.

PR #114, opened 2017-08-10
Updated badges.

jrbenny35:
As per title.


stephendonner:
r+ for this simple change; I see it working in https://github.com/jrbenny35/FoxPuppet/tree/update_badges

PR #123, opened 2017-09-08
Enable headless Firefox for beta and developer edition

davehunt:
Let's see if we can enable headless for beta and dev now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,15 +10,15 @@ _aliases:
       firefox: latest-nightly
     before_script: skip
   - &dev
-    env: FIREFOX=dev
+    env: FIREFOX=dev MOZ_HEADLESS=1
     addons:
-      <<: *apt
       firefox: latest-dev
+    before_script: skip
   - &beta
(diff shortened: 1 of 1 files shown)


davehunt:
Oops, we already tried this in #113. I'll be patient and wait for 57 to hit beta/dev.

davehunt:
hmm.. now nightly is failing on test_tracking_protection_shield which seems unrelated to this pull request. any thoughts @jrbenny35?

jrbenny35:
> hmm.. now nightly is failing on test_tracking_protection_shield which seems unrelated to this pull request. any thoughts

Looks like the page never loaded. Ran it again and it passed. 

I never liked the idea of using washingtonpost but it seems to work most of the time.

davehunt:
> Looks like the page never loaded. Ran it again and it passed.

I had assumed something more sinister as it was both the nightly jobs that failed. I've retriggered the build so we should hopefully be able to merge this. Once Firefox Quantum is released, we can make ESR the only exception to headless mode.

PR #133, opened 2017-11-28
Added the option to click on an installed extension from the navbar.

jrbenny35:
This should allow users to interact with an extension after they install it by simply knowing the name.

Since webextensions can not change the URL of the current page, FoxPuppet with switch to the new tab that gets opened once the extension is clicked. If no new tab is opened the current tab will still be selected for interaction.

I also moved some of the notification fixtures from the ```test_notifications.py``` file to the ```conftest.py``` file as I use them for this new test.


jrbenny35:
I have no idea why the flake8 build is failing. Any ideas?

davehunt:
> I have no idea why the flake8 build is failing. Any ideas?

It looks like flake8 is now checking files in the `.eggs` directory. I'm not sure why that's changed, but we could probably safely add this to the list of ignored paths.

jrbenny35:
> I don't see a compelling argument for this patch. If you want to interact with an extension button on the nav bar you can currently do this outside of FoxPuppet, where you can also cater for any consequences from clicking the button. This seems too restrictive to me, in that clicking a button must cause a non-specific change in the window handles, we switch to the last window handle (these are not guaranteed to be sorted in any particular order), and we assume the current URL before the click is 'about:blank'.

This is true to an extent. If an addon/extension has a URL it wants to navigate to it MUST do it within another tab. https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/create

Now 'about:blank' is the default URL that get's opened when a new tab is opened. Maybe it is too much to assume a new tab will ALWAYS be opened when you click on an extension. It would be better to check if a new tab is opened and if so, check that the URL changes.

> This seems too restrictive to me, in that clicking a button must cause a non-specific change in the window handles
I could change the click to not switch tabs and the user must do it manually.

davehunt:
> This is true to an extent. If an addon/extension has a URL it wants to navigate to it MUST do it within another tab.

But not all extensions will navigate to a URL when the button is clicked (think of the popups in Snooze Tabs or Containers).

> I could change the click to not switch tabs and the user must do it manually.

By manually you mean external to FoxPuppet? This would only leave locating/clicking of the button, which is also trivial to do outside of FoxPuppet.

I'd be okay adding a helper function for opening a tab in FoxPuppet, which could take a function as an argument. This helper would call the function and wait for the new tab to be opened, and switch to it. I can see this being useful in a few places.

whimboo:
I feel we are in a loop here, given that we have discussed this already on another PR. So grab what Puppeteer is using.

davehunt:
> I feel we are in a loop here, given that we have discussed this already on another PR. So grab what Puppeteer is using.

Sounds good to me. Let's close this pull request, and shift efforts to a helper for opening a tab. This could be added to #126, which can be reduced to just opening a tab via the new tab button, dropping the close/select tab parts for now.

PR #143, opened 2018-01-25
Bump geckodriver in Travis to latest, v0.19.1

stephendonner:
I confess I haven't-yet tested this locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,7 +91,7 @@ env:
   global:
     - DISPLAY=:99.0
 before_install:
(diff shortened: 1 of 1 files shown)


stephendonner:
Bah; I should've known it would bomb on the ESR builds :-(

davehunt:
> Bah; I should've known it would bomb on the ESR builds :-(

We coud probably just add a GECKODRIVER_VERSION environment variable, so each channel can specify what it uses. Let me see what I can do.

PR #144, opened 2018-01-26
Run everything headless except ESR

davehunt:


davehunt:
Thanks for the review @stephendonner. This is failing in Travis due to what seems to be a regression in Nightly. @jrbenny35 would you have time to investigate? It could be a genuine regression, or potentially just a locator change.

jrbenny35:
@davehunt I'll take a look today.

PR #151, opened 2018-02-23
Switch to Pipenv and fix a few isort issues

davehunt:


tests/test_notifications.py
```diff
@@ -3,9 +3,9 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 """Tests for the notifications API."""
 
+import pytest
```
jrbenny35: I wonder how these were missed??

davehunt: I think it's because we're now running flake8 with pytest installed, and it's able to correctly identify it as a third party package...

jrbenny35: Ah that makes sense.

PR #178, opened 2018-06-27
Bump Firefox ESR's geckodriver to latest, v. 0.21.0

stephendonner:


stephendonner:
No - not even close; sorry!

davehunt:
hmm.. I would have expected that to work.. we should also be able to make it headless now

davehunt:
It looks like it installed Firefox 52.9.0 via https://download.mozilla.org/?product=firefox-esr-latest&lang=en-US&os=linux64 so perhaps we need to wait longer before this is updated?

davehunt:
According to https://wiki.mozilla.org/Release_Management/Calendar the current ESR is 52.9 *and* 60.1 and we may need to wait until 2018-09-04 when it becomes 60.2.

stephendonner:
@davehunt https://www.mozilla.org/en-US/firefox/organizations/

screen shot 2018-06-27 at 5 54 35 am

It appears there are two branches?  One is "Firefox ESR 60" and the other is "Firefox 52 ESR" - not sure if those pretty-names matter, here?  (Assuming the placement of ESR differing is intentional.)

PR #181, opened 2018-07-20
Fix test by switching to Mozilla's tracking-protection test page

stephendonner:


stephendonner:
Well, this bombed all the way down! 👎

davehunt:
Did this work locally? Looking at the failing test, it's only on Nightly so I suspect that it's not related to the page we're loading. Also, the screenshot of the failure shows the tracking protection shield: https://s3.amazonaws.com/net-mozaws-stage-fx-test-travis-ci-artifacts/mozilla/FoxPuppet/1256/1256.3/py27.html. Perhaps @jrbenny35 mioght be able to help here? Maybe it's a locator change that landed in mozilla-central.

davehunt:
I wonder if the failure is related to https://bugzilla.mozilla.org/show_bug.cgi?id=1471713

davehunt:
If so, we could try changing the locator to `tracking-protection-icon-box` for Firefox >= 63, or setting `toolkit.cosmeticAnimations.enabled` to `False`.

davehunt:
Looks like the latter has helped. See #183. I'll leave this open in case there was another reason for the change.

davehunt:
Oops. Said I'd leave it open and then immediately closed it! 🙄

PR #187, opened 2018-07-28
Switch from pip to pipenv install tox-travis

stephendonner:


stephendonner:
Yeah, huge apologies.  I was tired and multi-tasking extra poorly when working on this.

I started with "just" doing ```- export PIPENV_IGNORE_VIRTUALENVS=1``` (which was in the wrong place, and non-functional anyways, as @jrbenny35 deftly pointed out).

I then noticed what I thought to be a legacy ```pip install tox-travis``` call, bowled over the bumper rails, and went into the other lane -- landing, finally, in the gutter.

Closing now, since I clearly over-simplified a task which didn't even need to be done.

PR #193, opened 2018-09-03
Sleep before the tracking protection test

davehunt:
Not ideal, but I'm not sure it's worth spending any longer investigating this.

### DIFF
--- a/tests/test_browser_model.py
+++ b/tests/test_browser_model.py
@@ -3,6 +3,8 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 """Tests for the browser model API."""
 
+import time
+
 import pytest
 from selenium.webdriver.support.wait import WebDriverWait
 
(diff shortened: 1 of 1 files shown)


tests/test_browser_model.py
```diff
@@ -49,6 +51,7 @@ def test_switch_to(foxpuppet, selenium):
 def test_tracking_protection_shield(foxpuppet, selenium):
     """Tests if the tracking protection icon displays."""
     browser = foxpuppet.browser
+    time.sleep(5)  # it seems necessary to sleep before navigating to get the shield
```
whimboo: Looks like they maybe fetch some data after first start, or update the database. Maybe just ask Johann Hofmann if he knows. If yes, there might be a way to circumvent this.

whimboo: Ah, and don't you use a call to wait here?

davehunt: Yes, this is my guess too. You'll notice that I closed this pull request though, as it didn't fix the issue. It looks like the 'regression' may be Linux only, so I'll dig into it a little more today.

whimboo: Ok, let me know.

Issue #194, opened 2018-09-04
Nightly and Developer Editions are failing in Travis CI


davehunt:
The Nightly builds have been failing in Travis CI since [12-08-2018](https://travis-ci.org/mozilla/FoxPuppet/builds/415006021), and Developer Edition has been failing since [29-08-2018](https://travis-ci.org/mozilla/FoxPuppet/builds/421855897). I was initially able to replicate locally (on macOS) but it turns out that might be an unrelated timing issue. It could be this 'regression' is only affecting Linux.

jrbenny35:
Has this been fixed?

davehunt:
Yes, thanks @jrbenny35, though I think I've seen some failures. Perhaps we have a flaky test. :/

PR #195, opened 2018-09-04
Update method of checking tracking protection in Firefox 63

davehunt:
Also documented some tips on keeping up with Firefox changes, and improved support for mozregression.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,27 +13,23 @@ _aliases:
     addons:
       <<: *artifacts
       firefox: latest-nightly
-    before_script: skip
   - &dev
     env: FIREFOX=dev
     addons:
       <<: *artifacts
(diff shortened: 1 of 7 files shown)


davehunt:
We can't merge this right now, because current ESR does not work well with https://itisatrap.org/firefox/its-a-tracker.html. This should be fixed within the next 24 hours because ESR 52 is due to be retired today.

davehunt:
Fixes #194

davehunt:
Yay, the tests pass! I've pushed a commit to fix the coverage, as we had some code paths for ESR 52 that we no longer need to support. Once Travis CI and coveralls is happy, would you mind reviewing @jrbenny35? Once merged, we should probably do a release, too.

PR #209, opened 2018-10-10
Fix addon complete notification

### LINKED ISSUE #208: Addon install confirmed notification has changed
In the latest nightly, the addon install notification has changed. It has moved to the app menu and, in turn, has a set of new class names and ID's.

ID: ```id="appMenu-addon-installed-notification"```.

Bugzilla [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1491438).

jrbenny35:
Fixes #208

### DIFF
--- a/foxpuppet/windows/browser/notifications/addons.py
+++ b/foxpuppet/windows/browser/notifications/addons.py
@@ -49,15 +49,18 @@ class AddOnInstallConfirmation(BaseNotification):
 
 class AddOnInstallComplete(BaseNotification):
     """Add-on install complete notification."""
+
     def close(self):
         """Close the notification."""
         with self.selenium.context(self.selenium.CONTEXT_CHROME):
             if self.window.firefox_version > 63:
                 self.root.find_anonymous_element_by_attribute(
-                    'anonid', 'button').click()
+                    "anonid", "button"
+                ).click()
                 self.window.wait_for_notification(None)
             else:
                 BaseNotification.close(self)
+
 
 class AddOnInstallRestart(BaseNotification):
     """Add-on install restart notification."""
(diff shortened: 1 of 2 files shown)


foxpuppet/windows/browser/notifications/addons.py
```diff
@@ -67,6 +78,7 @@ class AddOnProgress(BaseNotification):
     "addon-install-blocked-notification": AddOnInstallBlocked,
     "addon-install-confirmation-notification": AddOnInstallConfirmation,
     "addon-install-complete-notification": AddOnInstallComplete,
+    "appMenu-addon-installed-notification": AddOnInstallComplete,
```
davehunt: We should add a comment to indicate that we can clean up anything that not needed once ESR version is later than 63

foxpuppet/windows/browser/window.py
```diff
@@ -20,6 +20,10 @@ class BrowserWindow(BaseWindow):
     _file_menu_new_window_button_locator = (By.ID, "menu_newNavigator")
     _nav_bar_locator = (By.ID, "nav-bar")
     _notification_locator = (By.CSS_SELECTOR, "#notification-popup popupnotification")
+    _app_menu_notification_locator = (
+        By.CSS_SELECTOR,
+        "#appMenu-addon-installed-notification",
```
davehunt: This locator appears specific to the installed notification. Is this the only notification to appear within the app menu? Out of curiosity can you share a screenshot of what this looks like? If there are likely to be other notifications in this area, we should try to make this more generic.

jrbenny35: From my debugging there are other ones shown but not at the same time. Its the same area that shows the `Restart and Restore` for when firefox has an update. And they also seem to have different ID's.  That is also why I added  ```self.root.is_displayed()```.

Some examples: ```appMenu-update-restart-notification```, ```appMenu-update-manual-notification```.

[Screenshot](https://send.firefox.com/download/554609f68e/#vbXheWfPsKXHKtODy8ijHg)

davehunt: After doing a bit of investigation, it appears that we have a `<panel>` with id "appMenu-notification-popup". This has several child `<popupnotification>`s, only one of which is displayed. I would suggest that we could use the CSS locator `#appMenu-notification-popup popupnotification` and find the first one that is displayed. This would allow FoxPuppet to support the other app menu notifications in the future.

davehunt: Something like this should give you the first displayed notification:

```python
_app_menu_notification_locator = (By.CSS_SELECTOR, '#appMenu-notification-popup popupnotification')
notifications = self.selenium.find_elements(*_app_menu_notification_locator)
notification = next(n for n in notifications if n.is_displayed())
```

jrbenny35: I didn't put 2 and 2 together to realize that there are multiple elements with the same name... 😨

jrbenny35:
I now wonder if the ```NoSuchElement``` exception is needed, as it seems that the notification root is always even if there isn't a notification shown.

PR #243, opened 2019-02-11
Add support for new notification selectors

willdurand:
The notifications have been refactored in FF (see: https://phabricator.services.mozilla.com/D17699) and there is no anonymous element anymore. Instead, we can use `find_element` directly and use a `by.CLASS_NAME` selector.

This PR updates the notification code to be compatible with Nightly 67.0a1 and above, while keeping backward compatibility for older versions.

This patch has been tested locally on https://github.com/mozilla/addons-frontend.


jrbenny35:
Looks like there are some flake8 issues. Could you take a look at the travis build?

willdurand:
> Looks like there are some flake8 issues. Could you take a look at the travis build?

yes, sorry. I just reinstalled my `flake8` tox env locally.

PR #319, opened 2024-10-19
chore(project): move black to dev-dependencies

### LINKED ISSUE #315: Our pyproject file has the black dependency in the wrong group.
Currently `black` is located under `dependencies`, it should be under `dev.dependencies`.

Temidayo32:
chore(project): move black to dev-dependencies

Because

* black is a development tool and should not be part of production dependencies

This commit

* moves black from dependencies to dev-dependencies in the poetry configuration

Fixes #315


Temidayo32:
@b4handjr I'm not sure I understood perfectly what you meant when you said it merged with the wrong title. But Please review if I  didn't address the issue correctly.

b4handjr (approved):
Thanks!

PR #325, opened 2024-11-19
chore(project): Enforce code coverage

### LINKED ISSUE #323: Enforce code coverage
As this project grows, it will be great to use the pytest plugin `pytest-cov` to enforce code coverage. We will need to make a push to increase the coverage but adding a check to the CI runs to not allow a decrease in the current coverage would be great.

cashall-0:
Because

* We need to ensure that as the project grows, we push for increased coverage by adding a check to the CI runs to not allow a decrease in the current coverage.

This commit

* Enforces a 95.41% code test coverage

Fixes #323

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 BLACK_CHECK = black -l 90 --check --diff .
 BLACK_FIX = black -l 90 .
-FOXPUPPET_TESTS = pytest -vvv --driver Firefox --cov --html results/report.html
+MINIMUM_COVERAGE = 95.41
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,6 +1,7 @@
 BLACK_CHECK = black -l 90 --check --diff .
 BLACK_FIX = black -l 90 .
-FOXPUPPET_TESTS = pytest -vvv --driver Firefox --cov --html results/report.html
+MINIMUM_COVERAGE = 95.41
```
b4handjr, line 3: We can make this a nice round 95

b4handjr (changes requested):
Just 1 small nit

Makefile
cashall-0: done

b4handjr (approved):
Thanks for this!

PR #327, opened 2024-12-13
type(feat): Implement Bookmark Classes and Tests

### LINKED ISSUE #312: Add Bookmark Functionality
Add functionality to NavBar (and elsewhere) to set, unset, and retrieve bookmarks.

Refer to https://github.com/mozilla/fx-desktop-qa-automation/blob/main/modules/browser_object_navigation.py for ideas on what needs to be built and how to do it.

Add tests that exercise all methods. Make sure they run in `make test` and in CI.

Make sure the existing tests also pass using `make test` and in CI.

Temidayo32:
This PR add functionalities to set, unset, and retrieve bookmarks

Because

* Extend FoxPuppet by implementing the Bookmark functionality

This commit

* Add functionalities (classes and tests) to set, unset and retrieve bookmarks.

Fixes #312


Temidayo32:
@b4handjr Current tests didn't meet 95% requirement hence the failure. I will add missing tests to meet the requirements.

foxpuppet/windows/browser/bookmarks/bookmark.py
```diff
@@ -0,0 +1,227 @@
...
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            star_button_image = self.find_element(BookmarkLocators.STAR_BUTTON_IMAGE)
+            if star_button_image and star_button_image.get_attribute("starred") == "true":
+                self.click_element(BookmarkLocators.STAR_BUTTON)
+                self.click_element(BookmarkLocators.REMOVE_BUTTON)
+
+
+class AdvancedBookmark(BasicBookmark):
+    """Handles advanced bookmark operations."""
+
```
b4handjr, line 105: What's an "Advanced Bookmark" exactly?

Temidayo32: Okay, so in Desktop Smoke test, a similar term is used so I decided to adopt it in FoxPuppet. Basically, an advanced bookmark is a more detailed way of adding a bookmark, which involves `name`, `url`, `tags`, and `keywords`, as against just simply using the `url` alone when creating a `Basic Bookmark` with the `star button`.

b4handjr: Okay let me ask Ben C for clarifications

Temidayo32: Here is a link to it: https://github.com/mozilla/fx-desktop-qa-automation/blob/65fe54ea0badf6d665e55b4649da32c8a10b4598/modules/browser_object_navigation.py#L206

b4handjr: @ben-c-at-moz Can you explain what your "Advanced Bookmark" action does exactly?

Temidayo32: Yeah, I agree, the name is a little confusing.

ben-c-at-moz: But I'm fine if we just want to keep using it for more detailed bookmark operations, as long as we explain it.

ben-c-at-moz (commented):
I think this is good, but I'd like it to pass all checks before merging.

Temidayo32:
> I think this is good, but I'd like it to pass all checks before merging.

Actually, it's failing in Nightly build due to this: https://github.com/browser-actions/setup-firefox/issues/621

foxpuppet/windows/browser/bookmarks/bookmark.py
b4handjr: Could we just call it `Bookmarks` and put all the functionality into one class? If a user wants to do it through the "Other Bookmarks" option, we can just have a flag or something when they create the bookmark.

foxpuppet/windows/browser/bookmarks/bookmark.py
```diff
@@ -0,0 +1,218 @@
...
+            root (:py:class:`~selenium.webdriver.remote.webelement.WebElement`): WebDriver element object for bookmark
+
+        Returns:
+            :py:class:`BaseBookmark`: Bookmark instance or None
+        """
+        with window.selenium.context(window.selenium.CONTEXT_CHROME):
+            try:
+                return BasicBookmark(window, root)
+            except NoSuchElementException:
+                return None
```
b4handjr, line 44: I am curious as to why you'd like to return None and not throw the error.

b4handjr (changes requested):
Had a quick look but I think I want the bookmark functionality all in 1 class.

tests/test_bookmarks.py
```diff
@@ -0,0 +1,177 @@
...
+def advanced_bookmark(
+    browser: BrowserWindow, bookmark: Bookmark, selenium: WebDriver
+) -> AdvancedBookmark:
+    """Return a AdvancedBookmark instance."""
+    selenium.get(bookmark.url)
+    bookmark_instance = browser.wait_for_bookmark(AdvancedBookmark)
+    if bookmark_instance is None:
+        raise ValueError("Failed to get AdvancedBookmark instance.")
+    if not isinstance(bookmark_instance, AdvancedBookmark):
+        raise ValueError("Expected AdvancedBookmark instance but got different type")
```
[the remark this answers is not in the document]
Temidayo32: I didn't have those lines in initially. So if I remove those lines, mypy will raise type errors like this

```python

tests/test_bookmarks.py:95: error: Item "BasicBookmark" of "BasicBookmark | None" has no attribute
"add_bookmark"  [union-attr]
        bookmark_instance.add_bookmark(bookmark_data)
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
tests/test_bookmarks.py:95: error: Item "None" of "BasicBookmark | None" has no attribute "add_bookmark"
[union-attr]
        bookmark_instance.add_bookmark(bookmark_data)
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
tests/test_bookmarks.py:96: error: Incompatible return value type (got "BasicBookmark | None", expected
"AdvancedBookmark")  [return-value]
        return bookmark_instance
```

I will do a different work around for this when I refactor it.

foxpuppet/windows/browser/bookmarks/bookmark.py
Temidayo32: I think it might be less verbose. But, now again, it could confusing and unclear. I will just correct that.

foxpuppet/windows/browser/navbar.py
```diff
@@ -36,3 +48,67 @@ def is_tracking_shield_displayed(self) -> bool:
...
+            self.selenium.switch_to.default_content()
+
+    def open_main_menu(
+        self, locator_toolbar: Tuple[str, str], locator_menu_bar: Tuple[str, str]
+    ) -> None:
+        """Activate main menu bar"""
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            self.context_click(locator_toolbar)
+            self.click_element(locator_menu_bar)
+
```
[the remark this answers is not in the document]
b4handjr: Well some of these just seem like generic actions, like `switch_to_default_context()` and `find_element`.

tests/test_bookmarks.py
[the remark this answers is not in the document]
b4handjr: This is a good error, but the solution isn't what you currently have. Looking forward to seeing how you figure this out

foxpuppet/windows/browser/navbar.py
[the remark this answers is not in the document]
Temidayo32: Will you suggest I move them to a class higher up the hierarchy?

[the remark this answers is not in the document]
Temidayo32: So I moved all the generic actions into `region.py`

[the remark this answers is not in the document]
b4handjr: I think these should be removed if they aren't specifically related to bookmark functionality. We can add these later if need be.

Temidayo32:
> Let me know when you'd like me to look at this again.

Yeah, you can. Thanks.

b4handjr:
> > Let me know when you'd like me to look at this again.
> 
> Yeah, you can. Thanks.

Well I added a comment above too

foxpuppet/windows/browser/navbar.py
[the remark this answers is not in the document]
b4handjr: I understand but a lot of them are doing what selenium has built in. Selenium has a `find_element` method already, as an example.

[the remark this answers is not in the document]
Temidayo32: @b4handjr I have refactored and removed all of them. It is now ready for another review.

Temidayo32:
Hi @b4handjr The code is now ready for another review. Thanks.

foxpuppet/windows/browser/bookmarks/bookmark.py
```diff
@@ -0,0 +1,212 @@
...
+            )
+            return star_button_image.get_attribute("starred") == "true"
+
+    @property
+    def is_bookmarked_menu(self) -> bool:
+        """Checks if the current page is bookmarked using the main bookmark menu.
+
+        Returns:
+            bool: True if the page is bookmarked, False otherwise.
+        """
```
b4handjr, line 62: This is saying that if it's bookmarked via the menu it won't also have a star, but will only show in the menu? I don't quite understand this property.

foxpuppet/windows/browser/bookmarks/bookmark.py
```diff
@@ -0,0 +1,212 @@
...
+                menu_items = bookmark_menu.find_elements(
+                    By.CSS_SELECTOR, "menuitem.bookmark-item"
+                )
+                for item in menu_items:
+                    item_label = item.get_attribute("label")
+                    if item_label and item_label.lower() in current_page_title.lower():
+                        return True
+        return False
+
+    def add(self) -> None:
```
b4handjr, line 87: What's the difference with `add()` or `add_bookmark()`?

foxpuppet/windows/browser/bookmarks/bookmark.py
```diff
@@ -0,0 +1,212 @@
...
+
+            if bookmark_data["url"]:
+                self.actions.send_keys(bookmark_data["url"] + Keys.TAB).perform()
+
+            tags = bookmark_data["tags"]
+            if tags is not None:
+                for tag in tags:
+                    self.actions.send_keys(tag).perform()
+                    self.actions.send_keys(",").perform()
+                self.actions.send_keys(Keys.TAB).perform()
```
b4handjr, line 129: Think you could use the walrus operator here? https://peps.python.org/pep-0572/

foxpuppet/windows/browser/bookmarks/bookmark.py
```diff
@@ -0,0 +1,212 @@
...
+            label (str): The name of the bookmark to search for.
+        """
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            self.selenium.find_element(*BookmarkLocators.PANEL_MENU).click()
+            self.selenium.find_element(*BookmarkLocators.PANEL_BOOKMARK_MENU).click()
+            panel_bookmarks = self.selenium.find_element(
+                *BookmarkLocators.PANEL_BOOKMARK_TOOLBAR
+            )
+            if panel_bookmarks:
+                menu_items = panel_bookmarks.find_elements(
```
b4handjr, line 154: Why do you check if the `panel_bookmarks` is true? Wouldn't selenium throw an error if it couldn't find the element needed?

foxpuppet/windows/browser/bookmarks/bookmark.py
```diff
@@ -0,0 +1,212 @@
...
+                    By.CSS_SELECTOR, "toolbarbutton.bookmark-item"
+                )
+                if menu_items:
+                    for item in menu_items:
+                        item_label = item.get_attribute("label")
+                        if item_label and label.lower() in item_label.lower():
+                            return True
```
b4handjr, line 161: This could probably be done with a list comprehension.

foxpuppet/windows/browser/bookmarks/bookmark.py
```diff
@@ -0,0 +1,212 @@
...
+        return False
+
+    def delete(self) -> None:
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            star_button_image = self.selenium.find_element(
+                *BookmarkLocators.STAR_BUTTON_IMAGE
+            )
+            if star_button_image and star_button_image.get_attribute("starred") == "true":
+                self.selenium.find_element(*BookmarkLocators.STAR_BUTTON).click()
+                self.selenium.find_element(*BookmarkLocators.REMOVE_BUTTON).click()
```
b4handjr, line 171: Is this just for the individual page? Also, why a `delete()` and then a `delete_bookmark()`, we could have 1 method that takes a flag for how the user wants to delete the bookmark.

tests/test_bookmarks.py
```diff
@@ -0,0 +1,150 @@
...
+    @property
+    def keyword(self):
+        """Returns the current list of keywords."""
+        return self._keyword
+
+    @keyword.setter
+    def keyword(self, new_keyword: str):
+        """Sets new keywords for the bookmark."""
+        if isinstance(new_keyword, str):
+            self._keyword = new_keyword
```
b4handjr, line 33: With static typing we shouldn't need these type checks right?

tests/test_bookmarks.py
```diff
@@ -0,0 +1,150 @@
...
+def test_add_bookmark_star(
+    bookmark_star: Bookmark,
+) -> None:
+    """Ensure the bookmark is added using the star button."""
+    assert bookmark_star.is_bookmarked_star
+
+
+def test_add_bookmark_menu(bookmark_menu: Bookmark) -> None:
+    """Ensure the bookmark is added using the main bookmark menu."""
+    assert bookmark_menu.is_bookmarked_menu
```
b4handjr, line 111: The test is to "add" a bookmark but the API is `is_bookmarked_menu`, that is a bit confusing.

tests/test_bookmarks.py
```diff
@@ -0,0 +1,150 @@
...
+def test_delete_bookmark_star(bookmark_star: Bookmark) -> None:
+    """Delete bookmark added using star button."""
+    bookmark_star.delete()
+    assert bookmark_star.is_bookmarked_star is False
+
+
+def test_retrieve_bookmark_menu(bookmark_menu: Bookmark) -> None:
+    """Retrieve Bookmark added using the main bookmark menu."""
+    label = "Internet for people"
+    assert bookmark_menu.retrieve_bookmark(label) is True
```
b4handjr, line 129: Isn't this the same test as `test_retrieve_bookmark_star`?

tests/test_bookmarks.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def test_add_bookmark_menu(bookmark_menu: Bookmark) -> None:
+    """Ensure the bookmark is added using the main bookmark menu."""
+    assert bookmark_menu.is_bookmarked_menu
+
+
+def test_retrieve_bookmark_star(bookmark_star: Bookmark) -> None:
+    """Retrieve bookmark added using star button."""
+    label = "Internet for people"
```
b4handjr, line 116: This `label` variable could be a fixture as I see you use it often.

tests/test_bookmarks.py
```diff
@@ -0,0 +1,150 @@
...
+    bookmark_star.delete()
+    assert bookmark_star.retrieve_bookmark(label) is False
+
+
+def test_retrieve_deleted_bookmark_menu(bookmark_menu: Bookmark) -> None:
+    """Test retrieve deleted bookmark (main bookmark menu)."""
+    label = "Internet for people"
+    bookmark_menu.delete_bookmark(label)
+    assert bookmark_menu.retrieve_bookmark(label) is False
```
b4handjr, line 150: These assertions could be a part of the other delete tests.

tests/test_bookmarks.py
```diff
@@ -0,0 +1,150 @@
...
+    with pytest.raises(ValueError, match="Keyword must be a string."):
+        bookmark_entry.keyword = 123
+
+
+def test_invalid_tags(bookmark_entry: BookmarkEntry) -> None:
+    """Test setting a non-list or invalid list to tags raises ValueError."""
+    with pytest.raises(ValueError, match="Tags must be a list of strings."):
+        bookmark_entry.tags = "tag1, tag2"
+    with pytest.raises(ValueError, match="Tags must be a list of strings."):
+        bookmark_entry.tags = [123, "tag2"]
```
b4handjr, line 99: This can be 2 tests

tests/test_bookmarks.py
```diff
@@ -0,0 +1,150 @@
...
+        "keyword": bookmark_entry.keyword,
+    }
+    bookmark_using_menu.add_bookmark(bookmark_data)
+    return bookmark_using_menu
+
+
+def test_invalid_keyword(bookmark_entry: BookmarkEntry) -> None:
+    """Test setting a non-string value to the keyword raises ValueError."""
+    with pytest.raises(ValueError, match="Keyword must be a string."):
+        bookmark_entry.keyword = 123
```
b4handjr, line 91: I don't think we will need this test

tests/test_bookmarks.py
```diff
@@ -0,0 +1,150 @@
...
+        """Returns the current list of tags."""
+        return self._tags
+
+    @tags.setter
+    def tags(self, new_tags: list[str]):
+        """Sets new tags for the bookmark."""
+        if isinstance(new_tags, list) and all(isinstance(t, str) for t in new_tags):
+            self._tags = new_tags
+        else:
+            raise ValueError("Tags must be a list of strings.")
```
b4handjr, line 48: This seems like the wrong place to do this validation, shouldn't it be within the `tags` property? Also, shouldn't our static typing alert to an improperly set tag?

tests/test_bookmarks.py
```diff
@@ -0,0 +1,150 @@
...
+    def __init__(self, url: str, name: str = "", keyword: str = "", tags: list[str] = []):
+        self.url = url
+        self.name = name
+        self._keyword = keyword
+        self._tags = tags or []
+
+    @property
+    def keyword(self):
+        """Returns the current list of keywords."""
+        return self._keyword
```
b4handjr, line 27: Is this a list or just 1 word, it is currently set as a `str`.

foxpuppet/windows/browser/navbar.py
```diff
@@ -20,7 +21,10 @@ class NavBar(Region):
 
     """
 
-    _tracking_protection_shield_locator = (By.ID, "tracking-protection-icon-box")
+    _tracking_protection_shield_locator: Tuple[str, str] = (
+        By.ID,
+        "tracking-protection-icon-box",
+    )
```
b4handjr, line 27: Is this change needed?

foxpuppet/windows/browser/bookmarks/bookmark.py
```diff
@@ -0,0 +1,212 @@
...
+                self.actions.send_keys(Keys.TAB).perform()
+
+            if bookmark_data.get("keyword"):
+                keyword = bookmark_data["keyword"] or ""
+                self.actions.send_keys(keyword + Keys.TAB).perform()
+
+            self.actions.send_keys(Keys.TAB, Keys.TAB, Keys.TAB, Keys.ENTER).perform()
+            self.actions.send_keys(Keys.TAB, Keys.ENTER).perform()
+
+            self.selenium.switch_to.default_content()
```
b4handjr, line 138: Using the `with` clause, we don't need to switch back to default content here.

foxpuppet/windows/browser/bookmarks/bookmark.py
```diff
@@ -0,0 +1,212 @@
...
+
+            frame_element = self.selenium.find_element(
+                *BookmarkLocators.ADD_BOOKMARK_FRAME
+            )
+            if frame_element is not None:
+                self.selenium.switch_to.frame(frame_element)
+            else:
+                raise NoSuchElementException(
+                    f"iFrame with locator {BookmarkLocators.ADD_BOOKMARK_FRAME} not found."
+                )
```
b4handjr, line 115: Selenium will raise an error before the program gets here, if you want to raise an error with specific context you'll have to use a `try/except` block.

foxpuppet/region.py
```diff
@@ -6,7 +6,10 @@
 from selenium.webdriver.remote.webelement import WebElement
 from selenium.webdriver.support.wait import WebDriverWait
 from selenium.webdriver.remote.webdriver import WebDriver
+from selenium.webdriver.common.action_chains import ActionChains
+from selenium.common.exceptions import NoSuchElementException
 from foxpuppet.windows import BaseWindow
+from typing import Tuple
```
b4handjr, line 12: I don't this this is used in here?

foxpuppet/region.py
```diff
@@ -6,7 +6,10 @@
 from selenium.webdriver.remote.webelement import WebElement
 from selenium.webdriver.support.wait import WebDriverWait
 from selenium.webdriver.remote.webdriver import WebDriver
+from selenium.webdriver.common.action_chains import ActionChains
+from selenium.common.exceptions import NoSuchElementException
```
b4handjr, line 10: I don't this this is used in here?

b4handjr (changes requested):
Lots of great work here, this is really coming together! I would still like to see some consolidation and clarification with it comes to adding or deleting a bookmark. Also, there are some `isinstance` checks that I think aren't needed as we are using types to verify that.

foxpuppet/windows/browser/navbar.py
Temidayo32: Probably not

foxpuppet/windows/browser/bookmarks/bookmark.py
Temidayo32: I should probably use a flag instead and combine both into one method so it isn't confusing

tests/test_bookmarks.py
Temidayo32: I got too familiar with the code. I will improve on clarity.

tests/test_bookmarks.py
Temidayo32: Yeah, they should be about the same. The only difference is how bookmarks are added.

tests/test_bookmarks.py
Temidayo32: I will correct that clarity mistake.

foxpuppet/windows/browser/bookmarks/bookmark.py
Temidayo32: Okay, the reason is a result of the way Mozilla URL works. If I paste the Mozilla URL like this `https://www.mozilla.org`, the fully loaded page will be localized, as in, `https://www.mozilla.org/en-GB/`. So if I bookmark the page using the star, the localized URL is what gets bookmarked, not the root URL. However, when I bookmark using the main menu, I use the root URL, which won't match the localized Mozilla URL displayed in the browser, so it won't show as bookmarked If I check using the star button.

Temidayo32: I wasn't sure if using a localized Mozilla URL might be okay though. Could I just use a localized one?

foxpuppet/windows/browser/bookmarks/bookmark.py
Temidayo32: I thought to go with a generator expression here which might better fit, since the intention here is not to create a list, but to evaluate a condition.

foxpuppet/windows/browser/bookmarks/bookmark.py
ben-c-at-moz: For a test of this logic, a localized URL should be fine.

foxpuppet/windows/browser/bookmarks/bookmark.py
```diff
@@ -0,0 +1,213 @@
...
+                    self.selenium.find_element(*BookmarkLocators.MENU_BAR).click()
+                    self.selenium.find_element(
+                        *BookmarkLocators.MAIN_MENU_BOOKMARK
+                    ).click()
+                    self.actions.context_click(
+                        self.selenium.find_element(*BookmarkLocators.MANAGE_BOOKMARKS)
+                    ).perform()
+                    self.selenium.find_element(*BookmarkLocators.ADD_BOOKMARK).click()
+
+                    frame_element = self.selenium.find_element(
```
b4handjr, line 87: Could we name this `bookmark_frame` or something like that?

foxpuppet/windows/browser/bookmarks/bookmark.py
```diff
@@ -0,0 +1,213 @@
...
+                        *BookmarkLocators.ADD_BOOKMARK_FRAME
+                    )
+                    self.selenium.switch_to.frame(frame_element)
+                    if bookmark_data is not None:
+                        if bookmark_data["name"]:
```
b4handjr, line 92: I think you could use the walrus operator here, and it would allow `bookmark_data` to be used bellow as well.

foxpuppet/windows/browser/bookmarks/bookmark.py
```diff
@@ -0,0 +1,213 @@
...
+                            self.actions.send_keys(keyword + Keys.TAB).perform()
+
+                    self.actions.send_keys(
+                        Keys.TAB, Keys.TAB, Keys.TAB, Keys.ENTER
+                    ).perform()
+                    if (
+                        folder := self.selenium.find_element(
+                            *BookmarkLocators.BOOKMARK_FOLDER
+                        )
+                    ) is not None:
```
b4handjr, line 118: I don't think we need to do this `None` checks. You can remove the `is not None` part and do the same check as well. Also, good use of the walrus operator!

foxpuppet/windows/browser/bookmarks/bookmark.py
```diff
@@ -0,0 +1,213 @@
...
+                        folder.click()
+                        self.selenium.switch_to.frame(folder)
+                        self.actions.send_keys(Keys.TAB, Keys.ENTER).perform()
+
+    def retrieve_bookmark(self, label: str) -> bool:
+        """
+        Check if a bookmark with the given label exists.
+
```
b4handjr, line 126: The function name reads as if this will return a bookmark but it seems to be more of a check. Can you update the name to reflect that?

foxpuppet/windows/browser/bookmarks/bookmark.py
```diff
@@ -0,0 +1,213 @@
...
+                if (item_label := item.get_attribute("label"))
+            ):
+                return True
+        return False
+
+    def delete_bookmark(
+        self, label: Optional[str] = None, is_detailed: bool = False
+    ) -> None:
+        """
+        Delete a bookmark using either quick delete (star button) or detailed menu approach.
```
b4handjr, line 151: I wonder if this is overly complex? Does it matter how a bookmark is deleted?

foxpuppet/windows/browser/bookmarks/bookmark.py
```diff
@@ -0,0 +1,213 @@
...
+                    *BookmarkLocators.STAR_BUTTON_IMAGE
+                )
+                if (
+                    star_button_image
+                    and star_button_image.get_attribute("starred") == "true"
+                ):
+                    self.selenium.find_element(*BookmarkLocators.STAR_BUTTON).click()
+                    self.selenium.find_element(*BookmarkLocators.REMOVE_BUTTON).click()
+            else:
+                self.actions.context_click(
```
b4handjr, line 173: For operations like this where you have just 1 condition to check you can just have an `if` with the expression you want to evaluate ending with a `return` and then outside of that `if` block you can add the actions that will be done if the expression evaluates opposing to what you wanted.

So something like
```
if True:
    code here if true
    return
code here if false
```

Does that make sense?

foxpuppet/windows/browser/bookmarks/bookmark.py
```diff
@@ -0,0 +1,213 @@
...
+from typing import TYPE_CHECKING, Optional, TypedDict, List
+
+
+class BookmarkData(TypedDict):
+    """Bookmark properties."""
+
+    name: str
+    url: str
+    tags: Optional[List[str]]
+    keyword: Optional[str]
```
b4handjr, line 20: 👍🏾

b4handjr (changes requested):
Looking great, nice and pointed towards the bookmarks. Just had some follow up points I wanted to mention.

foxpuppet/windows/browser/bookmarks/bookmark.py
Temidayo32: The only reason I added the more detailed approach is because if we test adding a bookmark using the more detailed approach, then we should do same with the delete.

b4handjr (approved):
R+ great work as always!! 🎉 🚀

PR #330, opened 2025-01-11
Updated Setup-firefox action for FoxPuppet

### LINKED ISSUE #329: Roll our own setup-firefox action until the mainline version updates
Currently there is a failure on our CI workflow due to an error with the `setup-firefox` action that we use. The repo hasn't been updated in a while, so we should work on a solution until it does get an update.

We can take the fixes found here https://github.com/browser-actions/setup-firefox/pull/623 and build our own version of `setup-firefox` that we store in this repo until an upstream fix has been made. 

We will need to build an `index.js` file that has the fixes linked above and point our github actions workflow to it.

Temidayo32:
Because

* Our current setup-firefox build is causing tests to fail in CI. While we await fixes to be pushed in the setup-firefox repo, we can utilize a local build that includes the fix pulled from the repo (https://github.com/browser-actions/setup-firefox/pull/623)

This commit

* Adds a local setup-firefox build that can used in our GitHub actions for FoxPuppet

Fixes #329

### DIFF
--- a/.github/actions/setup_firefox/action.yml
+++ b/.github/actions/setup_firefox/action.yml
@@ -1,0 +1,14 @@
+name: 'Setup Firefox'
+description: 'Install and setup Firefox'
+author: "Shin'ya Ueoka"
+inputs:
+  firefox-version:
+    description: 'The Firefox version to install and use. Examples: 84.0, 84.0.1, latest-esr'
+outputs:
+  firefox-version:
+    description: 'The installed Firefox version. Useful when given a latest version.'
+  firefox-path:
+    description: 'The installed Firefox path.'
+runs:
+  using: 'node20'
+  main: 'index.js'
--- a/.github/actions/setup_firefox/index.js
+++ b/.github/actions/setup_firefox/index.js
@@ -1,0 +1,7985 @@
+/******/ (() => { // webpackBootstrap
+/******/ 	var __webpack_modules__ = ({
+
+/***/ 9484:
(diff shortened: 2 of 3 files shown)


b4handjr:
Thanks for working on this. Looks like the tests are still failing for the same reason. I think you need to point our workflow for the tests to use your custom workflow now.

Temidayo32:
> Thanks for working on this. Looks like the tests are still failing for the same reason. I think you need to point our workflow for the tests to use your custom workflow now.

Yeah, that's the next thing. But don't I need to modify our workflow in settings? which I don't have access to.

b4handjr:
> > Thanks for working on this. Looks like the tests are still failing for the same reason. I think you need to point our workflow for the tests to use your custom workflow now.
> 
> Yeah, that's the next thing. But don't I need to modify our workflow in settings? which I don't have access to.

No, you can edit the workflow here https://github.com/mozilla/FoxPuppet/blob/main/.github/workflows/linux_tests.yml

Temidayo32:
> > > Thanks for working on this. Looks like the tests are still failing for the same reason. I think you need to point our workflow for the tests to use your custom workflow now.
> > 
> > 
> > Yeah, that's the next thing. But don't I need to modify our workflow in settings? which I don't have access to.
> 
> No, you can edit the workflow here https://github.com/mozilla/FoxPuppet/blob/main/.github/workflows/linux_tests.yml

I didn't realize that. Let me do that right now

Temidayo32:
Great! the fix now works!

b4handjr (approved):
Nicely done!

PR #337, opened 2025-01-15
Github deployment workflow for PyPI

### LINKED ISSUE #333: Create github workflow to build and deploy to pypi

Temidayo32:
Because

* We need to start deploying FoxPuppet for external consumption by building and deploying to PyPI

This commit

* This PR adds a file for building, testing the build, and deploying the build to PyPI

Fixes #333

### DIFF
--- a/.github/workflows/foxpuppet_deploy.yml
+++ b/.github/workflows/foxpuppet_deploy.yml
@@ -1,9 +1,9 @@
 name: Publish to PyPI
 
 on:
-  pull_request:
-    branches:
-      - main
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
 
 env:
   PYTHON_VERSION: '3.12'
--- a/.github/workflows/foxpuppet_deploy.yml
+++ b/.github/workflows/foxpuppet_deploy.yml
@@ -28,7 +28,7 @@ jobs:
 
       - name: Setup Firefox
         id: setup-firefox
-        uses: ./.github/actions/setup_firefox/  
+        uses: ./.github/actions/setup_firefox/
         with:
           firefox-version: latest
 
@@ -45,16 +45,8 @@ jobs:
           python -m venv wheel-env
           source wheel-env/bin/activate
           pip install dist/*.whl
-          pip install pytest pytest-selenium pytest-cov pytest-html
           mkdir -p results
-          export PYTHONUNBUFFERED=1
-          export MOZ_HEADLESS=1
-          export GECKODRIVER_LOG="$(pwd)/results/geckodriver.log"
-          pytest -vvv \
-            --driver Firefox \
-            --capture=no \
-            --cov --cov-fail-under=95 \
-            --html results/report.html 2>&1 | tee results/pytest.log
+          make test
           deactivate
 
       - name: Install and Test the Tarball
@@ -64,14 +56,7 @@ jobs:
           pip install dist/*.tar.gz
           pip install pytest pytest-selenium pytest-cov pytest-html
           mkdir -p results
-          export PYTHONUNBUFFERED=1
-          export MOZ_HEADLESS=1
-          export GECKODRIVER_LOG="$(pwd)/results/geckodriver.log"
-          pytest -vvv \
-            --driver Firefox \
-            --capture=no \
-            --cov --cov-fail-under=95 \
-            --html results/report.html 2>&1 | tee results/pytest.log
+          make test
           deactivate
 
       - name: Publish to PyPI
--- a/.github/workflows/foxpuppet_deploy.yml
+++ b/.github/workflows/foxpuppet_deploy.yml
@@ -49,6 +49,16 @@ jobs:
           make test
           deactivate
 
+      - name: Upload Wheel Test Artifacts
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: wheel-test-results
+          path: |
+            results/report.html
+            results/geckodriver.log
+          retention-days: 14
+
       - name: Install and Test the Tarball
         run: |
           python -m venv tarball-env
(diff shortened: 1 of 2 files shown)


b4handjr (changes requested):
Great start, this should get us most of the way there. I'd like to see some changes.

1. Lets use this action https://github.com/pypa/gh-action-pypi-publish
2. I'd like to upload the built files as a tagged release to github, I found some examples using gh cli to do that, but we will need a github token, which I can provide.
3. I'd like to trigger this via a TAG. That tag would start this workflow and if completed will create a release on github with the built files.

We will also need to trigger this action before we merge it to ensure it is all working

.github/workflows/foxpuppet_deploy.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: |
+          # Install the built wheel
+          poetry add dist/*.whl
+          make test
+
+      - name: Install and Test the Tarball
+        run: |
+          # Install the built tarball
+          poetry add dist/*.tar.gz
+          make test
```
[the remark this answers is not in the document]
Temidayo32: Do you mean I run the full `make check` or you are implying the need for more specific tests for the tarball build?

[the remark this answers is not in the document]
b4handjr: Well I am saying we are doing 2 back to back poetry installs.. how are we sure that we are testing the current install when we run `make test`?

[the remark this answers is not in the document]
Temidayo32: Oh, got it. Perhaps, a better approach will be to run the test for each build in separate venv or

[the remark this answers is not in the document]
b4handjr: We could do that. We would need to upload the individual artifacts to a following job which would then do the release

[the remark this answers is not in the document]
Temidayo32: Alright. I have made that adjustment already, and I have attended to the requested changes you left above. I pushed already

Temidayo32:
> Have you tried getting this to run on Actions so we can see it working?

I was hoping to get the API token for the PYPi and Github steps before attempting that. Or no need for that yet?

b4handjr:
> > Have you tried getting this to run on Actions so we can see it working?
> 
> I was hoping to get the API token for the PYPi and Github steps before attempting that. Or no need for that yet?

I don't want to push to pypi yet, we need to make sure everything works before I add the token

Temidayo32:
> > > Have you tried getting this to run on Actions so we can see it working?
> > 
> > 
> > I was hoping to get the API token for the PYPi and Github steps before attempting that. Or no need for that yet?
> 
> I don't want to push to pypi yet, we need to make sure everything works before I add the token

Yeah, I am doing that at the moment. I am testing the action workflow.

b4handjr:
The error on the publish job seems to be that there is no browser/geckodriver installed to run the tests on.

Temidayo32:
I have now fixed the Github action up to the point of pushing to PyPI

.github/workflows/foxpuppet_deploy.yml
```diff
@@ -0,0 +1,88 @@
...
+          pip install pytest pytest-selenium pytest-cov pytest-html
+          mkdir -p results
+          export PYTHONUNBUFFERED=1
+          export MOZ_HEADLESS=1
+          export GECKODRIVER_LOG="$(pwd)/results/geckodriver.log"
+          pytest -vvv \
+            --driver Firefox \
+            --capture=no \
+            --cov --cov-fail-under=95 \
+            --html results/report.html 2>&1 | tee results/pytest.log
```
b4handjr, line 66: Looks like make test is working now too?

Temidayo32: Yeah, exactly. But I need to ignore the non-zero exit code that is making the other CI to fail.

Makefile
```diff
@@ -1,3 +1,7 @@
+export PYTHONUNBUFFERED = 1
```
b4handjr, line 1: I don't think we need this, I have a feeling this is not what solved our issue

Makefile
```diff
@@ -1,3 +1,7 @@
...
+export MOZ_HEADLESS = 1
+export GECKODRIVER_LOG = $(shell pwd)/results/geckodriver.log
```
b4handjr, line 3: If we're exporting these we should capture them as an artifact, it could help with future debugging.

b4handjr (changes requested):
Just a few final comments

b4handjr:
I think this is ready to go, lets finalize this and make sure it runs only on Tags, then we can create a Tag and get this released!

.github/workflows/foxpuppet_deploy.yml
```diff
@@ -1,9 +1,9 @@
 name: Publish to PyPI
 
 on:
-  pull_request:
-    branches:
-      - main
+  push:
+    tags:
+      - "v*"
```
b4handjr, line 6: Lets get this to match sematic versioning.

b4handjr (changes requested):

.github/workflows/foxpuppet_deploy.yml
```diff
@@ -1,9 +1,9 @@
 name: Publish to PyPI
 
 on:
-  pull_request:
-    branches:
-      - main
+  push:
+    tags:
+      - "v*.*.*"
```
b4handjr, line 6: Check these docs for more specifics https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#patterns-to-match-branches-and-tags

b4handjr: We probably want something like `'v[0-9]+.[0-9]+.[0-9]+'`

b4handjr (approved):
Looks good!!

PR #338, opened 2025-01-26
type(feat): Implemented PanelUI Classes and Tests

### LINKED ISSUE #313: Add PanelUI functionality
Add functionality, including new classes (potentially in new files) to automate the PanelUI (Hamburger menu).

1. Not everything needs to be automated, but we need Classes for the menu and at least placeholders for the menu items.
2. Please automate the New Tab and New Window options.
3. Please write a method to return whether or not an update badge appears on the menu icon

Write tests for all added functionality. Make sure they pass with `make test` and in UI.

Make sure existing tests pass with `make test` and in UI.

Temidayo32:
Because

* This commit implements PanelUI functionality

This commit

* Implemented PanelUI Class and Tests

Fixes #313

### DIFF
--- a/foxpuppet/windows/browser/navbar.py
+++ b/foxpuppet/windows/browser/navbar.py
@@ -39,3 +39,35 @@ class NavBar(Region):
                 return el.get_attribute("active") is not None
             el = self.root.find_element(By.ID, "tracking-protection-icon")
             return bool(el.get_attribute("state"))
+
+    def url_bar(self, links: list) -> list:
+        """Check if the provided links are present in the url bar's suggestions."""
+        urls = []
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            for link in links:
+                url_bar = self.selenium.find_element(*NavBarLocators.INPUT_FIELD)
+                url_bar.clear()
+                url_bar.send_keys(link)
+
+                self.wait.until(
+                    lambda _: self.selenium.find_elements(*NavBarLocators.SEARCH_RESULTS)
+                )
+
+                search_results = self.selenium.find_elements(
+                    *NavBarLocators.SEARCH_RESULT_ITEMS
+                )
+
+                for result in search_results:
+                    url_span = result.find_element(*NavBarLocators.SEARCH_RESULT_ITEM)
+                    if url_span.text in link:
+                        if len(url_span.text) != 0:
+                            urls.append(link)
+                            break
+        return urls
+
+
+class NavBarLocators:
+    INPUT_FIELD = (By.ID, "urlbar-input")
+    SEARCH_RESULTS = (By.ID, "urlbar-results")
+    SEARCH_RESULT_ITEM = (By.CSS_SELECTOR, "span.urlbarView-url")
+    SEARCH_RESULT_ITEMS = (By.CSS_SELECTOR, "div.urlbarView-row[role='presentation']")
--- a/foxpuppet/windows/browser/navbar.py
+++ b/foxpuppet/windows/browser/navbar.py
@@ -5,7 +5,7 @@
 
 from selenium.webdriver.common.by import By
 from foxpuppet.region import Region
-from foxpuppet.windows.base import BaseWindow
+from foxpuppet.windows.browser.urlbar import UrlBar
 
 
 class NavBar(Region):
@@ -40,34 +40,7 @@ class NavBar(Region):
             el = self.root.find_element(By.ID, "tracking-protection-icon")
             return bool(el.get_attribute("state"))
 
-    def url_bar(self, links: list) -> list:
-        """Check if the provided links are present in the url bar's suggestions."""
-        urls = []
-        with self.selenium.context(self.selenium.CONTEXT_CHROME):
-            for link in links:
-                url_bar = self.selenium.find_element(*NavBarLocators.INPUT_FIELD)
-                url_bar.clear()
-                url_bar.send_keys(link)
-
-                self.wait.until(
-                    lambda _: self.selenium.find_elements(*NavBarLocators.SEARCH_RESULTS)
-                )
-
-                search_results = self.selenium.find_elements(
-                    *NavBarLocators.SEARCH_RESULT_ITEMS
-                )
-
-                for result in search_results:
-                    url_span = result.find_element(*NavBarLocators.SEARCH_RESULT_ITEM)
-                    if url_span.text in link:
-                        if len(url_span.text) != 0:
-                            urls.append(link)
-                            break
-        return urls
-
-
-class NavBarLocators:
-    INPUT_FIELD = (By.ID, "urlbar-input")
-    SEARCH_RESULTS = (By.ID, "urlbar-results")
-    SEARCH_RESULT_ITEM = (By.CSS_SELECTOR, "span.urlbarView-url")
-    SEARCH_RESULT_ITEMS = (By.CSS_SELECTOR, "div.urlbarView-row[role='presentation']")
+    @property
+    def url_bar(self) -> UrlBar:
+        """Returns an instance of the UrlBar class."""
+        return self._url_bar
--- a/foxpuppet/windows/browser/navbar.py
+++ b/foxpuppet/windows/browser/navbar.py
@@ -43,4 +43,4 @@ class NavBar(Region):
     @property
     def url_bar(self) -> UrlBar:
         """Returns an instance of the UrlBar class."""
-        return self._url_bar
+        return UrlBar(self.window, self.root)
--- a/foxpuppet/windows/browser/panel_ui/panel_ui.py
+++ b/foxpuppet/windows/browser/panel_ui/panel_ui.py
@@ -1,0 +1,225 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+"""Contains classes for handling Firefox Panel UI (Hamburger menu)."""
+
+from abc import ABCMeta
+
+from selenium.webdriver.common.by import By
+
+from foxpuppet.region import Region
+from selenium.webdriver.remote.webelement import WebElement
+from typing import Type, Any, TYPE_CHECKING, Optional
+
+
+class PanelUI(Region):
+    """Handles interaction with Panel UI."""
+
+    __metaclass__ = ABCMeta
+    if TYPE_CHECKING:
+        from foxpuppet.windows import BrowserWindow
+
+    @staticmethod
+    def create(
+        window: Optional["BrowserWindow"], root: WebElement
+    ) -> Type["PanelUI"] | Any:
+        """Create a Panel UI object.
+
+        Args:
+            window (:py:class:`BrowserWindow`): Window object this region
+                appears in.
+            root
+                (:py:class:`~selenium.webdriver.remote.webelement.WebElement`):
+                WebDriver element object that serves as the root for the
+                Panel UI.
+
+        Returns:
+            :py:class:`PanelUI`: Firefox Panel UI.
+
+        """
+        panel_items: dict = {}
+        _id: str | bool | WebElement | dict = root.get_property("id")
+
+        panel_items.update(PANEL_ITEMS)
+        return panel_items.get(_id, PanelUI)(window, root)
+
+    def open_panel_menu(self) -> None:
+        """
+        Opens the Panel UI menu.
+        """
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            self.selenium.find_element(*PanelUILocators.PANEL_UI_BUTTON).click()
+
+    def open_private_window(self) -> None:
+        """
+        Opens a new window in private browsing mode using the Panel UI menu.
+        """
+        # self.open_panel_menu()
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            self.selenium.find_element(*PanelUILocators.PRIVATE_WINDOW).click()
+
+    def open_history_menu(self) -> None:
+        """
+        Opens the History in Panel UI Menu
+        """
+        # self.open_panel_menu()
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            self.selenium.find_element(*PanelUILocators.HISTORY).click()
+
+    def wait_for_num_windows_or_tabs(self, expected_count: int) -> None:
+        """
+        Waits until the number of open browser windows or tabs matches the expected value.
+
+        Args:
+            expected_count (int): The expected number of windows or tabs.
+        """
+        self.wait.until(
+            lambda _: len(self.driver.window_handles) == expected_count,
+            f"Expected {expected_count} windows or tabs, but found {len(self.driver.window_handles)}",
+        )
+
+    def switch_to_new_window_or_tab(self) -> None:
+        """Get list of all window handles, switch to the newly opened tab/window"""
+        handles = self.selenium.window_handles
+        self.selenium.switch_to.window(handles[-1])
+
+
+class History(PanelUI):
+    """Handles interactions with Firefox History."""
+
+    def is_present(self, link: str) -> bool:
+        """
+        Checks if a specific link is present in the recent history.
+        Args:
+            link `str`: The URL or part of the URL to check for in the recent history.
+
+        Returns:
+            bool: `True` if the link is present in the recent history, `False` otherwise.
+        """
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            history_items = self.selenium.find_elements(
+                *PanelUILocators.RECENT_HISTORY_ITEMS
+            )
+            for item in history_items:
+                item_src = item.get_attribute("image")
+                if item_src and link in item_src:
+                    print(item_src)
+                    return True
+            return False
+
+    def clear_history(self):
+        """
+        Clears the browsing history.
+        """
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            self.selenium.find_element(*PanelUILocators.CLEAR_RECENT_HISTORY).click()
+            self.selenium.switch_to.frame(
+                self.selenium.find_element(*PanelUILocators.HISTORY_IFRAME)
+            )
+            with self.selenium.context(self.selenium.CONTEXT_CONTENT):
+                self.selenium.find_element(*PanelUILocators.DROPDOWN_HISTORY).click()
+                self.selenium.find_element(
+                    *PanelUILocators.CLEAR_HISTORY_EVERYTHING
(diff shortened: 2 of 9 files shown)


Temidayo32:
This is the WIP. Tests are failing because they do not meet 95% coverage yet.

foxpuppet/windows/browser/panel_ui/panel_ui.py
```diff
@@ -0,0 +1,288 @@
...
+            self.selenium.get(link)
+
+        self.selenium.switch_to.window(initial_window_handle)
+        self.open_panel_menu()
+        self.open_history_menu()
+        for link in links:
+            if self.is_present(link):
+                invalid_links.append(link)
+        return invalid_links
+
```
b4handjr, line 262: What's the logic behind these additions? These seem like actions that a user would do. Are you expecting a user to supply a list of URLs for these methods to verify? Why wouldn't they just do that themselves?

b4handjr (changes requested):

foxpuppet/windows/browser/panel_ui/panel_ui.py
Temidayo32: Yeah, you are right. The actions mirror what a user would do manually, but given that this is a test automation tool, I think its purpose is always to automate manual user actions. Technically, this isn't user functionality - it's testing infrastructure. In this sense, I try to automate certain expected behaviors of Private Window browsing vs Normal Window browsing, history etc.

Temidayo32: Yeah, the user will have to supply a list of URLs

b4handjr: Remember, we are just trying to programmatically represent a way for a user to interact with Firefox. We don't want to assume what combinations they will do, just provide them a way to do them.

Temidayo32: Okay,  that makes good sense. I will move those methods to tests. So they serve as part of FoxPuppet test suite. Will that be good or I just remove them totally.

Temidayo32: Desktop test  has similar tests: https://github.com/mozilla/fx-desktop-qa-automation/blob/main/tests/security_and_privacy/test_private_session_awesome_bar_exclusion.py

b4handjr (changes requested):
Heading in the right direction with this. I think there will need to be some sort of `History` class so that we can contain the History specific interactions in one place.

Temidayo32:
> Heading in the right direction with this. I think there will need to be some sort of `History` class so that we can contain the History specific interactions in one place.

Yes, I tried to do that initially. But this was difficult using the same `property(history)` and `wait_for_history` method defined on the `BrowserWindow` class as we did with others, especially using the `Panel-UI`, which in the dom is a list and all items are visible. Suppose we decided to extend later on by adding more list items like `PrivateWindow` or `Download`. In that case, it will prove difficult to interact with a specific one since they will all be visible.

foxpuppet/windows/browser/window.py
```diff
@@ -129,6 +143,36 @@ def wait_for_bookmark(self) -> Bookmark:
...
 
+    def wait_for_panel_ui(
+        self, panel_ui_class: Optional[Type[P]] = PanelUI  # type: ignore
+    ) -> Optional[P]:
+        """Wait for the specified PanelUI to be displayed.
+
+        Args:
+            panel_ui_class (:py:class:`PanelUI`, optional):
+                The PanelUI subclass to wait for. If `None` is specified, it
+                will wait for any panel UI to be displayed. Defaults to `PanelUI`.
```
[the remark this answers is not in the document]
Temidayo32: I forgot to adjust this docstring. If you check my initial commit. I add two subclasses of the `PanelUI` class: `PrivateWindow` and `History` subclasses. However, I found it difficult to handle both because both elements are simultaneously present and interactable once the panel-ui is opened. The `panel_ui` property, just like `notification` property defined on the `BrowserWindow` depend on just one of its items or classes interactable at any given point. If more than one is interactable, then it is difficult as the code will interact with the first available item even if that's not the element we want to interact with. I have struggled trying to work out another logic to bypass this problem, but I wasn't able to make a breakthrough. This is why I reverted to the one class `PanelUI` to hold all methods for interacting with its child elements. This easily bypass the problem.

foxpuppet/windows/browser/window.py
```diff
@@ -129,6 +143,36 @@ def wait_for_bookmark(self) -> Bookmark:
             )
             return self.bookmark
 
+    def wait_for_panel_ui(
```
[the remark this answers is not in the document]
Temidayo32: This is the actual correct docstring:
`Wait for the specified PanelUI item to be displayed.`
That was misleading. There is no reason to wait for panel_ui as it is visible once the browser is opened.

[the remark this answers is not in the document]
Temidayo32: PanelUI item, here, referring to menu items in the Panel Ui dropdown. History, for instance.

foxpuppet/windows/browser/navbar.py
```diff
@@ -39,3 +39,35 @@ def is_tracking_shield_displayed(self) -> bool:
...
+                search_results = self.selenium.find_elements(
+                    *NavBarLocators.SEARCH_RESULT_ITEMS
+                )
+
+                for result in search_results:
+                    url_span = result.find_element(*NavBarLocators.SEARCH_RESULT_ITEM)
+                    if url_span.text in link:
+                        if len(url_span.text) != 0:
+                            urls.append(link)
+                            break
```
b4handjr, line 65: This should be a separate method. If you need to create a `UrlBar` class to house it, that is fine, but this `url_bar` should be a property that just returns a class.

foxpuppet/windows/browser/panel_ui/panel_ui.py
```diff
@@ -43,6 +40,20 @@ def create(
         panel_items.update(PANEL_ITEMS)
         return panel_items.get(_id, PanelUI)(window, root)
 
+    @property
+    def is_barged(self) -> bool:
```
b4handjr, line 44: I can see where you get this name from but would a user understand what `barged` means? Honestly, if I didn't see the `get_attribute("barged")` I wouldn't know what it meant.

foxpuppet/windows/browser/window.py
```diff
@@ -223,3 +233,21 @@ def open_window(self, private: bool = False) -> Union["BrowserWindow", Any]:
...
+        self.wait.until(
+            lambda _: len(self.selenium.window_handles) == expected_count,
+            f"Expected {expected_count} windows or tabs, but found {len(self.selenium.window_handles)}",
+        )
+        return True
+
+    def switch_to_new_window_or_tab(self) -> None:
+        """Get list of all window handles, switch to the newly opened tab/window"""
+        handles = self.selenium.window_handles
+        self.selenium.switch_to.window(handles[-1])
```
b4handjr, line 253: I still struggle with if this is needed. These are things that aren't specific to FoxPuppet, so are they actually needed here?

tests/test_panel_ui.py
```diff
@@ -26,75 +27,147 @@ def panel_ui(browser: BrowserWindow) -> PanelUI | None:
     Returns:
         :py:class:`PanelUI`: FoxPuppet Panel UI object
     """
-    return browser.wait_for_panel_ui(PanelUI)
+    return browser.wait_for_panel(PanelUI)
 
 
-def test_open_new_tab(panel_ui: PanelUI) -> None:
+@pytest.fixture
+def history(panel_ui: PanelUI, browser: BrowserWindow) -> History | None:
```
b4handjr, line 34: This could be `browser_history`

tests/test_panel_ui.py
```diff
@@ -4,9 +4,10 @@
...
 
 import pytest
+import time
 from selenium.webdriver.remote.webdriver import WebDriver
 from foxpuppet.windows import BrowserWindow
-from foxpuppet.windows.browser.panel_ui.panel_ui import PanelUI
+from foxpuppet.windows.browser.panel_ui.panel_ui import PanelUI, History
 
 
 links = [
```
b4handjr, line 13: This could be a fixture

tests/test_panel_ui.py
```diff
@@ -26,75 +27,147 @@ def panel_ui(browser: BrowserWindow) -> PanelUI | None:
...
+    urls = []
+    panel_ui.open_new_tab()
+    browser.switch_to_new_window_or_tab()
+    for link in links:
+        selenium.get(link)
+    panel_ui.open_panel_menu()
+    panel_ui.open_history_menu()
+    for link in links:
+        if history.is_present(link):
+            urls.append(link)
```
b4handjr, line 85: You could do this in a list comprehension so it's all on 1 line

b4handjr (changes requested):
Excellent progress. I see some tests here that aren't quite testing the functionality of FoxPuppet but more so the functionality of Firefox. Can you identify which tests those are?

foxpuppet/windows/browser/window.py
Temidayo32: I added them as helper functions, but I can just cut them out.

b4handjr (changes requested):
Almost there. I think we can remove some checks and functions but there are also a few methods that we need to make sure they're doing what they say they will do.

foxpuppet/windows/browser/panel_ui/panel_ui.py
```diff
@@ -0,0 +1,187 @@
...
+                lambda _: set(self.selenium.window_handles) - initial_handles,
+                message="New window did not open",
+            )
+
+    def open_private_window(self) -> None:
+        """
+        Opens a new window in private browsing mode using the Panel UI menu.
+        """
+        initial_handles = set(self.selenium.window_handles)
+        self.open_panel_menu()
```
b4handjr, line 100: Cant we check that the window is private too?

foxpuppet/windows/browser/panel_ui/panel_ui.py
```diff
@@ -0,0 +1,187 @@
...
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            self.selenium.find_element(*PanelUILocators.NEW_WINDOW).click()
+            self.wait.until(
+                lambda _: set(self.selenium.window_handles) - initial_handles,
+                message="New window did not open",
+            )
+
```
b4handjr, line 94: I wonder if we should switch the selenium execution to the newly opened window or tab. Would be easy to do. What do you think?

foxpuppet/windows/browser/panel_ui/panel_ui.py
```diff
@@ -0,0 +1,187 @@
...
+        Returns:
+            list[WebElement]: List of WebElement objects representing history items.
+                Returns an empty list if no history items are found.
+        """
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            history_items = self.selenium.find_elements(
+                *PanelUILocators.RECENT_HISTORY_ITEMS
+            )
+            print(history_items)
```
b4handjr, line 135: Should this `print` be here?

foxpuppet/windows/browser/urlbar.py
```diff
@@ -0,0 +1,53 @@
...
+            search_results = self.selenium.find_elements(
+                *URLBarLocators.SEARCH_RESULT_ITEMS
+            )
+
+            suggested_urls = [
+                result.find_element(*URLBarLocators.SEARCH_RESULT_ITEM).text
+                for result in search_results
+                if result.find_element(*URLBarLocators.SEARCH_RESULT_ITEM).text
+            ]
+            print(suggested_urls)
```
b4handjr, line 44: Is this `print` supposed to be here?

foxpuppet/windows/browser/urlbar.py
```diff
@@ -0,0 +1,53 @@
...
+
+from selenium.webdriver.common.by import By
+from selenium.webdriver.remote.webdriver import WebDriver
+from selenium.webdriver.support.wait import WebDriverWait
+
+
+class UrlBar:
+    def __init__(self, selenium: WebDriver, wait: WebDriverWait):
+        self.selenium = selenium
+        self.wait = wait
```
b4handjr, line 14: Shouldn't this be a `Region` class?

foxpuppet/windows/browser/panel_ui/panel_ui.py
```diff
@@ -0,0 +1,187 @@
...
+        Checks if the Panel UI button indicates a pending Firefox update.
+
+        Returns:
+            bool: True if an update notification (barge) is present, False otherwise.
+        """
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            barged_status = self.selenium.find_element(
+                *PanelUILocators.PANEL_UI_BUTTON
+            ).get_attribute("barged")
+            return barged_status == "true"
```
b4handjr, line 56: I know this is called `barged` but lets call it `update_status`. That will make more sense in the future.

b4handjr (changes requested):
Looking good. I think we're almost there.

foxpuppet/windows/browser/panel_ui/panel_ui.py
```diff
@@ -104,6 +108,17 @@ def open_private_window(self) -> None:
                 lambda _: set(self.selenium.window_handles) - initial_handles,
                 message="Private window did not open",
             )
+            new_private_window = (
+                set(self.selenium.window_handles) - initial_handles
+            ).pop()
+            self.selenium.switch_to.window(new_private_window)
+
+            from foxpuppet.windows.browser.window import BrowserWindow
```
b4handjr, line 116: ehh, I don't quite like this even though I know why you're doing it. Let's just keep the imports at the top if we can.

foxpuppet/windows/browser/panel_ui/panel_ui.py
```diff
@@ -104,6 +108,17 @@ def open_private_window(self) -> None:
...
+
+            new_window = BrowserWindow(self.selenium, new_private_window)
+
+            if not new_window.is_private:
+                raise ValueError("The new window is not private.")
```
b4handjr, line 121: `ValueError` doesn't seem like the right choice here. Should we just raise an `Exception` with this message?

b4handjr (changes requested):
Nice changes! Just a couple things and ill have Ben C look at this too.

foxpuppet/windows/browser/panel_ui/panel_ui.py
Temidayo32: This will cause a circular import error.

b4handjr: Ignore me then

pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+addopts = -vvv --driver Firefox --cov --cov-fail-under=95 --html=results/report.html --self-contained-html
+testpaths = tests
```
b4handjr, line 3: Please add a newline

b4handjr (changes requested):
Adding these based on our chat

foxpuppet/windows/browser/panel_ui/panel_ui.py
```diff
@@ -0,0 +1,201 @@
...
+        with self.selenium.context(self.selenium.CONTEXT_CHROME):
+            history_items = self.selenium.find_elements(
+                *PanelUILocators.RECENT_HISTORY_ITEMS
+            )
+            return history_items
+
+    def clear_history(self):
+        """
+        Clears the browsing history.
+        """
```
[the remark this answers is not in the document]
Temidayo32: Yes, the `history_items` method

tests/test_panel_ui.py
```diff
@@ -0,0 +1,161 @@
...
+                    break
+        assert len(found_urls) == 3
+
+
+def test_verify_links_open_in_new_window_from_history(
+    panel_ui: PanelUI, browser_history: History, selenium: WebDriver, links: list
+) -> None:
+    """Test that links opened in new window are present in browser history."""
+    panel_ui.open_new_window()
+    time.sleep(3)
```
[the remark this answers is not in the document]
ben-c-at-moz: I think you could too, even the blank page has something you could wait for, iirc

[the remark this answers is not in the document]
ben-c-at-moz: Or you could even do this:
* Get the `len()` of `selenium.window_handles` (call it `old_num_windows` or something)
* open new window
* using `Wait()`, do `wait.until(lambda _: len(selenium.window_handles) < old_num_windows)

Not entirely sure exactly this would work, but something like it probably will

ben-c-at-moz (commented):
This all looks good, I think addressing that sleep command is the only new thing I can find in here that I think needs to change. I would definitely make absolutely sure you don't run `find_elements()` on anything until you KNOW it exists, as it doesn't include the same implicit wait as `find_element()` does.

tests/test_panel_ui.py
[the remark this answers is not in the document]
Temidayo32: I had to add this in the `open_new_window()` method.

```
self.wait.until(
                lambda _: self.selenium.execute_script("return document.readyState") == "complete",
                message="New window document not fully loaded",
            )

```

foxpuppet/windows/browser/panel_ui/panel_ui.py
```diff
@@ -95,6 +95,11 @@ def open_new_window(self) -> None:
             )
             new_window = (set(self.selenium.window_handles) - initial_handles).pop()
             self.selenium.switch_to.window(new_window)
+            self.wait.until(
+                lambda _: self.selenium.execute_script("return document.readyState")
+                == "complete",
+                message="New window document not fully loaded",
+            )
```
b4handjr, line 102: Is this needed?

Temidayo32: Yes, this more accurately addresses the root issue compared to `time.sleep`

tests/test_panel_ui.py
```diff
@@ -0,0 +1,159 @@
...
+
+def test_clear_recent_history(
+    panel_ui: PanelUI, browser_history: History, selenium: WebDriver
+) -> None:
+    """Test clearing browser history removes visited URLs."""
+    url = "https://www.mozilla.org/en-US/?v=a"
+    selenium.get(url)
+    panel_ui.open_history_menu()
+    browser_history.clear_history()
+    time.sleep(1)
```
b4handjr, line 150: Do you still need this sleep?

b4handjr (commented):
! final thing and I think this is good

b4handjr (approved):
Looks great!! Thanks for working on this and sticking with it, this is a great addition and will go a long way to helping us expand FoxPuppets offerings!

PR #343, opened 2025-01-31
type(chore): Fix github release action

### LINKED ISSUE #342: Fix github release action
During our last release, our Github action failed due to it not finding the files needed to add to the release. Lets either do another build or share the files with that job after we push to pypi.

We need to upload the built artifacts so that we can then download them in the github release job. You can use `upload-artifact` and `download-artifact` to do this. See https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow

Temidayo32:
Because

* Reason

This commit fixes Github release actions

* Fix Github release action.

Fixes #342

### DIFF
--- a/.github/workflows/foxpuppet_deploy.yml
+++ b/.github/workflows/foxpuppet_deploy.yml
@@ -79,6 +79,12 @@ jobs:
             results/geckodriver.log
           retention-days: 14
 
+      - name: Upload Package Artifacts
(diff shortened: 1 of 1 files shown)


b4handjr:
This looks good, can you fix the name of this PR though so that it conforms with the PR naming guidelines?

b4handjr (approved):

PR #345, opened 2025-02-05
type(fix): Ensure GitHub Action runs the latest versions of Firefox

### LINKED ISSUE #344: Ensure Firefox Nightly Updates to the latest in GitHub Actions Workflow
Currently, we don't have the latest Firefox Nightly running in GitHub action workflow. This should be fixed such that the latest Firefox Nightly is used when running github actions in CI

Temidayo32:
Because

* FoxPuppet needs to be updated to reflect new updates to the Firefox browser. This won't be possible if tests are not running again the latest versions of Firefox.
* The is_private property defined on the BrowserWindow class is outdated.

This commit

* Ensure FoxPuppet tests are running against the latest Firefox versions.
* Updates the is_private method to use Private Window ESM module. This is done using dynamic imports.

Fixes #344, #346

### DIFF
--- a/foxpuppet/windows/browser/bookmarks/bookmark.py
+++ b/foxpuppet/windows/browser/bookmarks/bookmark.py
@@ -4,7 +4,8 @@
 """Contains classes for handling Firefox bookmarks."""
 
 from selenium.webdriver.common.by import By
-from selenium.common.exceptions import NoSuchElementException
+from selenium.webdriver.support import expected_conditions as EC
+from selenium.webdriver.support.ui import WebDriverWait
 from selenium.webdriver.remote.webelement import WebElement
 from selenium.webdriver.common.keys import Keys
 from foxpuppet.windows.browser.navbar import NavBar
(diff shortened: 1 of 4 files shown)


foxpuppet/windows/browser/bookmarks/bookmark.py
```diff
@@ -197,7 +201,7 @@ class BookmarkLocators:
     MANAGE_BOOKMARKS = (By.ID, "bookmarksShowAll")
     MENU_BAR = (By.ID, "toggle_toolbar-menubar")
     NAME_FIELD = (By.ID, "editBMPanel_namePicker")
-    NAVIGATOR_TOOLBOX = (By.ID, "navigator-toolbox")
+    NAVIGATOR_TOOLBOX = (By.ID, "TabsToolbar")
```
b4handjr, line 204: That looks like the right one!

b4handjr (approved):
Looks good, thanks for getting this fixed!

PR #353, opened 2025-02-20
fix(project): Bookmark Selector Fix

### LINKED ISSUE #352: Fix Bookmark Star Button Selector
Given the recent failure in Firefox latest, there is a need to fix the bookmark star selector.

Temidayo32:
Because

* The current Bookmark selector for the star button causes failure in Firefox latest. The selector needs to be updated.

This commit

* Fixes Bookmark selector to match recent Firefox changes.

Fixes #352


b4handjr:
Could you add some more detail to the `Because` bullet point?

b4handjr (approved):
Nice simple fix that works for all of the browsers!

